PR #544, opened 2022-03-18
feat: Oracle Reputation PVE integration

joshocalico:
This is an effective port of [PR 393 in AAVE-UI](https://github.com/aave/aave-ui/pull/393).

We've thrown together a preview page for you guys so you can see it in action.
https://aave-interface-with-pve-qp51q5ug4-reputation.vercel.app
(Go to Ethereum Mainnet market; select a reserve)
|Light|Dark|
|:-:|:-:|
|Screen Shot 2022-03-18 at 8 58 28 pm|Screen Shot 2022-03-18 at 8 58 35 pm|

If videos are your style I give a guide to the changes here.
https://www.loom.com/share/3e0eef7b9beb4453907731e6c96b2837

Big changes are we're using your theme data to theme the widget.
Got any alterations? Feel free to reach out.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,6 +47,7 @@
     "@mui/icons-material": "^5.5.1",
     "@mui/material": "^5.5.1",
     "@paraswap/sdk": "5.6.0-alpha.3",
+    "@reputation.link/vyps-kit": "^0.5.7",
     "@visx/axis": "^2.6.0",
     "@visx/curve": "^2.1.0",
     "@visx/event": "^2.6.0",
--- a/pages/reserve-overview.page.tsx
+++ b/pages/reserve-overview.page.tsx
@@ -1,3 +1,4 @@
+import { ChainId } from '@aave/contract-helpers';
 import { Trans } from '@lingui/macro';
 import {
   Box,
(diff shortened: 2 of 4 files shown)


sakulstra:
@joshocalico thanks for checking out the codebase! And sorry for never answering in the original pr.

I just had a look and personally am a bit on the fence regarding the pr.
First of there's some technical stuff.
- It doesn't reliably work for all reserves and i personally think it's a bit weird that the api is based on token symbol instead of asset address making this a bit arbitrary imo. Here an example of a non working reserve `reserve-overview/?underlyingAsset=0xd2eec91055f07fe24c9ccb25828ecfefd4be0c41&marketName=amm_mainnet`
- I think aave interface should reduce dependencies not increase the. Having a dependency for what essentially is a link seems wrong to me
- I don't like that this integration is so prominent as I think it's not adding a lot of value (just my very personal opinion)

All of that said, i personally could imagine a HAL like integration, but i think [similar as HAL](https://snapshot.org/#/aave.eth/proposal/0xa730caeec3c28e014ff456b454186ef41c6c1f382cf0a7caa3d99c4ae16c8318) this should go trough snapshot governance to let the community decide.

PR #560, opened 2022-03-21
fix: added manifest, and fixed reconnection to gnosis wallet

sendra:


src/libs/web3-data-provider/Web3Provider.tsx
```diff
@@ -144,14 +144,19 @@ export const Web3ContextProvider: React.FC<{ children: ReactElement }> = ({ chil
 
   // first, try connecting to a gnosis safe
   useEffect(() => {
+    console.log('tried safe::: ', triedSafe);
```
sakulstra, line 147:
```suggestion
```

src/libs/web3-data-provider/Web3Provider.tsx
```diff
@@ -149,9 +149,13 @@ export const Web3ContextProvider: React.FC<{ children: ReactElement }> = ({ chil
...
         if (loadedInSafe) {
-          connectWallet(WalletType.GNOSIS).catch(() => {
-            setTriedSafe(true);
-          });
+          connectWallet(WalletType.GNOSIS)
+            .then(() => {
+              setTriedSafe(true);
+            })
+            .catch(() => {
+              setTriedSafe(true);
```
foodaka, line 157: this also true?

sendra: yes, its to store that it has tried to connect to gnosis. even if it fails (which will happen when you are not accessing the app through gnosis safe)

foodaka (approved):

PR #561, opened 2022-03-21
Fix/total dashboard apys

defispartan:
Currently the earned and debt apys are listed as a fraction of the net worth, which is confusing for users since the supply/debt APY for assets doesn't match the top line value in dashboard. This changes earned APY as a fraction of totalLiquidity and debt APY as a fraction of totalDebt.

Also added a tooltip to explain how this value is calculated


src/modules/dashboard/DashboardTopPanel.tsx
```diff
@@ -110,7 +110,11 @@ export const DashboardTopPanel = () => {
         <TopInfoPanelItem icon={<NetAPYIcon />} title={<Trans>Net APY</Trans>} loading={loading}>
           {currentAccount ? (
             <FormattedNumber
-              value={(user?.earnedAPY || 0) - (user?.debtAPY || 0)}
+              value={
+                (user?.earnedAPY || 0) *
```
sakulstra, line 114: Does it make sense to use big number here instead? Probably doesn't matter much but I fear there might be an overflow in certain Situations

sakulstra (changes requested):
Thanks for your first contribution to the new ui old fren <3

src/modules/dashboard/DashboardTopPanel.tsx
defispartan: Since the string is already coming in as a floating point I don't think it makes sense to convert to integer -> BigNumber

I did add a divide by zero check that was missing though :sweat_smile:

defispartan: Yeah that would be way cleaner

sakulstra (approved):

PR #570, opened 2022-03-24
Feat/repay with dust warning

defispartan:
When doing a max repay where the wallet balance can't cover the full debt amount and remaining debt is < $5, display the remaining debt amount in red with a warning tooltip

Old:

Screenshot from 2022-03-24 10-30-52

New: 

Screenshot from 2022-03-24 10-16-16

Screenshot from 2022-03-24 10-16-23


foodaka (dismissed):

foodaka:
@defispartan thanks for the detailed PR description, very good for reviewers

src/components/TextWithTooltip.tsx
```diff
@@ -11,12 +11,14 @@ export interface TextWithTooltipProps extends TypographyProps {
...
   // eslint-disable-next-line
   children?: ReactElement<any, string | JSXElementConstructor<any>>;
+  color?: string;
 }
 
 export const TextWithTooltip = ({
   text,
   icon,
   iconSize = 14,
+  color = 'text.muted',
```
sakulstra, line 21: Should this perhaps just consistently always be like that? I mean instead of making it a prop

sakulstra: Feels weird to me that in this case it should be error color while essentially it's just an info like in all other cases.

src/components/transactions/FlowCommons/TxModalDetails.tsx
```diff
@@ -92,13 +93,18 @@ export const DetailsNumberLineWithSub = ({
   amount,
   amountUSD,
   hideSymbolSuffix,
+  dustRemainingWarning,
```
sakulstra, line 96: Better give it a text props or sth similar generic as this will open a can of worms.
It's about making the text red, not about dust.

src/components/transactions/Repay/RepayModalContent.tsx
```diff
@@ -153,6 +154,11 @@ export const RepayModalContent = ({
     .multipliedBy(marketReferencePriceInUsd)
     .shiftedBy(-USD_DECIMALS);
 
+  const maxRepayWithDustRemaining =
+    isMaxSelected &&
+    displayAmountAfterRepayInUsd.toNumber() > 0 &&
+    displayAmountAfterRepayInUsd.toNumber() < 5;
```
sakulstra, line 160: Why the lower 5?

src/components/TextWithTooltip.tsx
defispartan: Reverted it back to the original. I thought it would make it stand out more but I agree it looked kinda weird

src/components/transactions/FlowCommons/TxModalDetails.tsx
defispartan: Switched it to a text prop called color

src/components/transactions/Repay/RepayModalContent.tsx
defispartan: The tooltip message seems odd unless you're actually leaving a small amount behind. Like if I have 10000 DAI debt and repay my max wallet balance of 50 DAI, having a warning for leaving dust doesn't seem relevant

I can remove it though if you think otherwise

src/components/transactions/FlowCommons/TxModalDetails.tsx
```diff
@@ -92,13 +93,14 @@ export const DetailsNumberLineWithSub = ({
...
   hideSymbolSuffix,
+  color,
 }: DetailsNumberLineWithSubProps) => {
   return (
     <Row caption={description} captionVariant="description" mb={4} align="flex-start">
       <Box sx={{ display: 'flex', flexDirection: 'column', alignItems: 'flex-end' }}>
         <Box sx={{ display: 'flex', alignItems: 'center' }}>
           <TokenIcon symbol={symbol} sx={{ mr: 1, fontSize: '16px' }} />
-          <FormattedNumber value={amount} variant="secondary14" />
+          <FormattedNumber value={amount} variant="secondary14" color={color ? color : undefined} />
```
sakulstra, line 103: No need for the ternary as it will be undefined anyways if not set

src/components/transactions/Repay/RepayModalContent.tsx
sakulstra: I would remove it as the 5 is somewhat arbitrary (5 one tokens are basically dust). It doesn't hurt to show it imo.

sakulstra (commented):
Lgtm besides the few minor notes

src/components/transactions/FlowCommons/TxModalDetails.tsx
defispartan: I have a ternary obsession haha

sakulstra (approved):

PR #571, opened 2022-03-24
Feat/net apy tooltip

defispartan:
Add explanation for net APY


foodaka (approved):

src/modules/dashboard/DashboardTopPanel.tsx
```diff
@@ -107,7 +108,16 @@ export const DashboardTopPanel = () => {
           )}
         </TopInfoPanelItem>
 
-        <TopInfoPanelItem icon={<NetAPYIcon />} title={<Trans>Net APY</Trans>} loading={loading}>
+        <TopInfoPanelItem
+          icon={<NetAPYIcon />}
+          title={
+            <div style={{ display: 'flex' }}>
```
sakulstra, line 114: Not at the PC right now so little hard to check, but I think we have a predefined textwithtooltip component somewhere

defispartan: That component was for wrapping the actual tooltip text. There was another component which took text and tooltip as props called ListTopInfoItem but that was specific to the dashboard headers

defispartan:
I reworded the tooltip to better explain this scenario:

Screenshot from 2022-03-24 14-35-59

The net APY is relative to *net worth*, so it's possible to have a lower net APY than debt APY if the user is leveraged

src/components/infoTooltips/NetAPYTooltip.tsx
```diff
@@ -0,0 +1,15 @@
+import { Trans } from '@lingui/macro';
+
+import { TextWithTooltip, TextWithTooltipProps } from '../TextWithTooltip';
+
+export const NetAPYTooltip = ({ ...rest }: TextWithTooltipProps) => {
+  return (
+    <TextWithTooltip {...rest}>
+      <Trans>
+        The net APY is the combined effect of all supply and borrow positions on net worth,
```
foodaka, line 9:
```suggestion
       Net APY is the combined effect of all supply and borrow positions on net worth,
```

foodaka (approved):

PR #577, opened 2022-03-25
Feat/add emode params

defispartan:
Add emode max ltv, liquidation thresold, and liquidation bonus to reserve overview for supported assets


sakulstra (approved):
Very cool, thx :+1:

sakulstra (approved):

PR #582, opened 2022-03-27
feat: add kovan v3 market config

defispartan:


sendra:
We should not have this market. As kovan will be deprecated in the near future, so it doesnt make sense to mantain it on the client

PR #589, opened 2022-03-28
fix: removed gas widget. and gasPrice calcs

sendra:
On this pr, I removed the custom gas calculation and gas widget selector, as to delegate gas price calculation to the wallet provider. As now a days it is much better optimized there.

So now the gas widget will only show a gas estimation so the user still sees and approximation of what the transaction will cost


sakulstra (approved):

MareskoZ:
QC passed

sakulstra (approved):

foodaka (approved):

PR #594, opened 2022-03-28
fix: use paraswap even when in fork

sendra:


src/hooks/useSwap.ts
```diff
@@ -48,7 +50,8 @@ const MESSAGE_MAP = {
 };
 
 export const useSwap = ({ swapIn, swapOut, variant, userId, max, chainId }: UseSwapProps) => {
-  const paraSwap = getParaswap(chainId);
+  const { currentChainId } = useProtocolDataContext();
```
sakulstra, line 53: hmm, an easier(imho more readable) fix would have been to just take `currentNetworkConfig.baseChainId || currentNetworkConfig.chainId`

sendra: Updated it to use underlyingChainId, as its the one that user sets up when adding fork

sakulstra (approved):

PR #597, opened 2022-03-29
Feat/restyle repay warning

defispartan:


src/components/transactions/FlowCommons/TxModalDetails.tsx
```diff
@@ -83,6 +83,8 @@ interface DetailsNumberLineWithSubProps {
   symbol: string;
   amount: string;
   amountUSD: string;
+  previousAmount?: string;
```
sakulstra, line 86: overall lgtm, but might make sense to have consistent naming with `DetailsNumberLine` where it's `value, futureValue` vs `amount, previousAmount`. I don't have an opinion on which naming is better.

Also not 100% sure if both components are needed, might make sense to combine them

sakulstra (dismissed):

src/components/transactions/FlowCommons/TxModalDetails.tsx
defispartan: Updated params to `value` and `futureValue`

PR #601, opened 2022-03-29
fix: added tooltip for emode on overview

sendra:


src/components/infoTooltips/EModeTooltip.tsx
```diff
@@ -0,0 +1,22 @@
+import { Trans } from '@lingui/macro';
+import { Link } from '../primitives/Link';
+
+import { TextWithTooltip, TextWithTooltipProps } from '../TextWithTooltip';
+
+export const EModeTooltip = ({ ...rest }: TextWithTooltipProps) => {
+  return (
+    <TextWithTooltip {...rest}>
+      <Trans>
+        E-Mode increases your LTV for a selected category of assets up to 97%.{' '}
```
sakulstra, line 10: does it make sense that this is hardcoded 97%? On the asset details we should be able to just use the actual ltv from the asset in emode. Alternatively we could remove the "up to 97%".

sakulstra (dismissed):

src/components/infoTooltips/EModeTooltip.tsx
sendra: Added your suggestion, and using now emode ltv from reserve

sakulstra (approved):

PR #616, opened 2022-03-31
Fix assets sorting by symbol

Jeiwan:
This PR:
1. Fixes assets sorting by symbol in Firefox. Currently, it doesn't work. For some reason, this code behaves differently in Chrome/Safari and Firefox:
    ```js
    ["DAI", "EURS", "USDC", "USDT", "sUSD", "AAVE", "RAI", "xSUSHI", "UNI"].sort(function(a, b ) { return b.toUpperCase() < a.toUpperCase() ? -1 : 0;})
    ```
    To make it work in Firefox, I changed `0` to `1`. Chrome/Safari are not affected.
1. Fixes the order of sorting by symbol: ascending order is A to Z, descending is Z to A. Currently, it's the opposite.

### DIFF
--- a/src/modules/markets/AssetsList.tsx
+++ b/src/modules/markets/AssetsList.tsx
@@ -40,7 +40,7 @@ export default function AssetsList() {
 
   if (sortDesc) {
     if (sortName === 'symbol') {
-      filteredData.sort((a, b) => (b.symbol.toUpperCase() < a.symbol.toUpperCase() ? -1 : 1));
+      filteredData.sort((a, b) => (a.symbol.toUpperCase() < b.symbol.toUpperCase() ? -1 : 1));
     } else {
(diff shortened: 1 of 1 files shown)


sakulstra (approved):
Good catch & thx for your contribution the reason for it working is that Firefox sort algorithm is stable where the chrome one isn't.

Jeiwan:
The list of markets on Ethereum is quite long and finding an asset is problematic when they're not sorted. I use the list to learn about parameters and utilization of different assets.

sakulstra:
@Jeiwan thx for the feedback! Yep on eth mainnet i can fully understand the point.
Currently the default sorting is first stables then alphabetic, but when there's more feedback like this we might wanna reconsider that default.

PR #636, opened 2022-04-04
Feat/update faq links

defispartan:


src/components/infoTooltips/EModeTooltip.tsx
```diff
@@ -20,7 +20,7 @@ export const EModeTooltip = ({
         />
         .{' '}
         <Link
-          href="https://docs.aave.com/faq/"
+          href="https://docs.aave.com/faq/aave-v3-features"
```
foodaka, line 23: Do you think we should take the user directly to the section? https://docs.aave.com/faq/aave-v3-features#high-efficiency-mode-e-mode

PR #656, opened 2022-04-12
feat: underline active page

defispartan:
Add underline to active page in navbar


sendra (approved):

MareskoZ:
@sendra QA passed

defispartan:
> @defispartan pushed the suggested solution in https://github.com/aave/interface/pull/661/files as it was easier than doing a comment. I think the solution is nicer as it's a generic way to add an `active` class to links pointing to the current route for all links. Also it seems like oleks already added the styles in here: https://github.com/aave/interface/pull/656/files#diff-0bb4e8916774abdd0e57ec15465883766049631721cec9450ff7b571f4ef676aR63 which will now work again.

Cool, I like that approach better, I'll close this one

PR #657, opened 2022-04-12
Feat/change default screen

defispartan:
Switch default screen to markets instead of dashboard


defispartan:
@sakulstra @sendra What was the reasoning for defaulting to the dashboard?

Showing market data is probably more desirable to users than a connect wallet modal which they can also already from the navbar. Especially on mobile it's kind of clunky to click top right -> markets to get off of a blank screen

sakulstra:
@defispartan this was changed around ~4 times now.
Perhaps we just need to admit that people have different preferences.

Iirc the last argument was (@iamanastasia correct me if i'm wrong) that we want people to use the app and they use the app on dashboard. Most people are not really interested in the reserves but only in the yield which they will see on the dashboard as well.

I think sth like https://github.com/aave/interface/issues/660 might make more sense - although #660 specifically could be fixed my actively linking to the market page instead of the dashboard.

Anyhow if we really want to change the behavior we might want to  keep the dashboard default (as changing the urls now will breaks links all over the place), but redirect to markets when the user has not signed in before(afaik if he has we store the address in localstorage)? Personally not a fan of this either as when linking to the dashboard we now "automatically" redirect to the markets page which is unintuitive.

defispartan:
> @defispartan this was changed around ~4 times now. Perhaps we just need to admit that people have different preferences.
> 
> Iirc the last argument was (@iamanastasia correct me if i'm wrong) that we want people to use the app and they use the app on dashboard. Most people are not really interested in the reserves but only in the yield which they will see on the dashboard as well.
> 
> I think sth like #660 might make more sense - although #660 specifically could be fixed by actively linking to the market page instead of the dashboard.
> 
> Anyhow if we really want to change the behavior we might want to keep the dashboard default (as changing the urls now will breaks links all over the place), but redirect to markets when the user has not signed in before(afaik if he has we store the address in localstorage)? Personally not a fan of this either as when linking to the dashboard we now "automatically" redirect to the markets page which is unintuitive.

Opened a discussion for this: https://github.com/aave/interface/discussions/696

PR #665, opened 2022-04-14
feat: add block explorer link for reserves

defispartan:
Add block explorer link to underlying asset next to asset name

Also move oracle link to the bottom row


MareskoZ:
@defispartan for mobile version those functionality wasn't added, if it's expected, QC Passed

sendra (approved):

PR #680, opened 2022-04-18
fix: enable wrapped base assets to be fauceted

defispartan:


sendra:
This should not be merged. As wrapped assets are not backed by native assets, so it will break functionality. Thats why it was not added from the start

PR #709, opened 2022-04-25
feat: special asset labels

defispartan:
Variety of small changes:

- Display - instead of 0 in Total Borrow column for assets with borrowing disabled (AAVE, stETH, etc.)
- Add warning message about rebasing to AMPL asset page
- On ETH mainnet, replace the borrow button on AAVE reserve page with a button and message about Staking
- Add a min height to reserve overview page for better display of assets without many rows (AAVE)


foodaka:
closing out

PR #710, opened 2022-04-26
fix: calculation of max withdrawable amount

defispartan:
Currently when the max withdraw button is pressed, the amount is calculated by requiring that collateral * (liquidationThreshold + cushion) is remaining. This results in users being forced to keep a large amount of collateral to cover a small borrow position as shown below.

This changes the calculation to require a minimum health factor of 1.07 (to be consistent with borrow logic), which allows users to withdraw a greater potion of their collateral (in scenario below user could now withdraw up to 14998.6 DAI).

withdrawIssueStatus

withdrawIssueModal

### DIFF
--- a/src/components/transactions/Withdraw/WithdrawModalContent.tsx
+++ b/src/components/transactions/Withdraw/WithdrawModalContent.tsx
@@ -75,7 +75,9 @@ export const WithdrawModalContent = ({
     const maxSelected = value === '-1';
     amountRef.current = maxSelected ? maxAmountToWithdraw.toString(10) : value;
     setAmount(value);
-    if (maxSelected) {
+    if (maxSelected && maxAmountToWithdraw.eq(underlyingBalance)) {
+      setWithdrawMax('-1');
+    } else {
       setWithdrawMax(maxAmountToWithdraw.toString(10));
     }
   };
--- a/src/components/transactions/Withdraw/WithdrawModalContent.tsx
+++ b/src/components/transactions/Withdraw/WithdrawModalContent.tsx
@@ -55,15 +55,12 @@ export const WithdrawModalContent = ({
     poolReserve.usageAsCollateralEnabled &&
     user.totalBorrowsMarketReferenceCurrency !== '0'
   ) {
-    // if we have any borrowings we should check how much we can withdraw without liquidation
-    // with 0.5% gap to avoid reverting of tx
-    const excessHF = valueToBigNumber(user.healthFactor).minus('1');
+    // if we have any borrowings we should check how much we can withdraw to a minimum HF of 1.07
+    const excessHF = valueToBigNumber(user.healthFactor).minus('1.07');
     if (excessHF.gt('0')) {
       maxCollateralToWithdrawInETH = excessHF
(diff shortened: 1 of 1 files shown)


MareskoZ:
QC Passed

src/components/transactions/Withdraw/WithdrawModalContent.tsx
```diff
@@ -55,15 +55,12 @@ export const WithdrawModalContent = ({
     poolReserve.usageAsCollateralEnabled &&
     user.totalBorrowsMarketReferenceCurrency !== '0'
   ) {
-    // if we have any borrowings we should check how much we can withdraw without liquidation
-    // with 0.5% gap to avoid reverting of tx
-    const excessHF = valueToBigNumber(user.healthFactor).minus('1');
+    // if we have any borrowings we should check how much we can withdraw to a minimum HF of 1.07
+    const excessHF = valueToBigNumber(user.healthFactor).minus('1.07');
```
sakulstra, line 59: I'm not sure if I'm a big.fan of that - if you got a big balance in volatile assets 0.07hf swings are quite possible so I'd fear people get liquidated quite easily.rwflecting badly on the protocol. Wouldn't it be better to be a bit more konservative here? I understand the point if the borrow side is mostly dust, if it's not though this can get painful fast.

sendra: Maybe this should be accompanied with a check that users should check acknowledging that they are going to be near liquidation. This could appear only on certain conditions. But generally speaking I agree with @sakulstra here. It is dangerous to generically let users withdraw so close to 1 as depending on position it can supper easily get liquidated.

The question here, is if this safe gards should be managed by the interface, or if all responsability should be delegated to the user. in my opinion provably the best option is to let users withdraw until hf = 1 but if the case (we can put even a range to hf=1.07 or whatever) and add a check so user aknowledges the risk

defispartan: > I'm not sure if I'm a big.fan of that - if you got a big balance in volatile assets 0.07hf swings are quite possible so I'd fear people get liquidated quite easily.rwflecting badly on the protocol. Wouldn't it be better to be a bit more konservative here? I understand the point if the borrow side is mostly dust, if it's not though this can get painful fast.

The same logic for limiting users to a minimum of 1.07hf already exists for borrowing and as far as I know this hasn't caused any issues. 

I agree with @sendra that adding a checkbox for users to acknowledge the risk of borrow/withdraw below a certain threshold (1.2 maybe), but since this will require a design I think we should move forward with this because there are multiple users complaining about this issue weekly.

sakulstra: Okay fine for me. Agree that it doesn't make sense to have a double standard. Would highly recommend though to at least show a warning message when future hf is below 1.2 if we don't already do.

defispartan: There's a warning message for < 1.5 in both borrow and withdraw: "Liquidation risk is high. Lower amounts recomended."

sakulstra (approved):

PR #727, opened 2022-04-29
feat: risk checkbox

defispartan:
When borrowing or withdrawing below a HF of 1.5: 
- Display warning message and checkbox for users to acknowledge risk
- Disable borrow or withdraw button if checkbox is displayed and not accepted

Also reduced the minimum HF on withdraw from 1.07 -> 1.01. The reasoning for this is that if a user has a HF below 1.07 and can't repay from their wallet balance or collateral, then they are effectively locked into their position with the 1.07 limit.


defispartan:
@iamanastasia

src/components/transactions/Borrow/BorrowModalContent.tsx
```diff
@@ -141,6 +142,7 @@ export const BorrowModalContent = ({
     ),
     currentLiquidationThreshold: user.currentLiquidationThreshold,
   });
+  const displayRiskCheckbox = newHealthFactor.toNumber() < 1.5;
```
sakulstra, line 145: Doesn't this need to also check for `!= -1`?

cypress/support/steps/main.steps.ts
```diff
@@ -75,11 +74,13 @@ export const borrow = (
     amount,
     apyType,
     hasApproval = true,
+    isRisk = false,
```
foodaka, line 77: what is isRisk?

cypress/support/steps/main.steps.ts
```diff
@@ -115,8 +116,14 @@ export const borrow = (
     it(`Borrow ${amount} amount for ${_shortName}`, () => {
       setAmount({
         amount,
-        hasApproval,
       });
+    });
+    if (isRisk) {
+      it(`Click risk checkbox`, () => {
```
foodaka, line 122: a risk checkbox

cypress/support/steps/main.steps.ts
foodaka: maybe rename this to something like acceptsLiquidationRisk ?

PR #755, opened 2022-05-09
fix: update to aave utility version which fixes flash repay

sakulstra:
- https://github.com/aave/aave-utilities/pull/342


defispartan:
Closing since we discussed this isn't an issue

PR #770, opened 2022-05-11
fix: fix repay with collateral

sakulstra:
Resolves some issues with the current repay with collateral implementation:
- Excludes pools which are returning 500 errors for swaps involving certain assets (LINK, TUSD)
- Improve slippage estimation for input amounts


sakulstra:
@defispartan there still seems to be an issue - which weirdly i can only reproduce on ethereum where the flashAmount for some reason is smaller then then maxAmountToPay (reproducible via tenderly).

No idea where it's coming from exactly.

defispartan:
I think the same logic is needed for price impact as well. One thing I didn't know until researching further is that price impact and slippage are not the same thing: https://doc.paraswap.network/price-impact-slippage.

When you test a swap with high slippage, you can see that the expected input amount is calculated *before* price impact is taken into account:

rwc

sirsoth:
Im able to replicate errors on a position with LINK as collateral, unable to repay USDT or DAI

sirsoth:
Just Error getting txParams

foodaka:
is this also related to this issue https://github.com/aave/interface/issues/786 ?

sirsoth:
Looks to me to be the same error

foodaka (approved):

foodaka:
Thanks @defispartan for investigating and running this down

PR #788, opened 2022-05-16
Ambire Wallet batch transaction integration

0xdex18:
This PR implements batching for ERC20 approvals with Ambire, which is a smart wallet that allows for transaction batching, smart recovery, gas abstractions and others. Rather than having to sign & send the ERC20 approval transaction separately, with this PR, it will be batched together with the specific action when the wallet used is Ambire.

The UX impact is immense, as the approval step is essentially permanently eliminated, without compromising security (the same calls are being done, just in one tx).

While this PR implements some logic custom to Ambire, this logic can be easily modified into an implementation of the batching EIP that’s currently being worked on by Gnosis Safe and Ambire, turning the code into a general implementation of txn batching.

This implementation mainly adds the function sendBatchTx, similar to sendTx, that uses WalletConnect. Also, the PR for [aave-utilities](https://github.com/aave/aave-utilities/pull/348) is required to properly complete this PR.
At the moment, a npm version of modified aave-utilities is used to make this PR work but will have to be changed with the official npm package of @aave/contracts-helper, that adds an argument to skip the gas estimation (that would make for example the swap tx gas estimation fail if the approval is not done yet)

Please advise if there's a better way

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "test:open": "DOTENV_CONFIG_PATH='.env.local' cypress open"
   },
   "dependencies": {
-    "@aave/contract-helpers": "1.2.1",
+    "@aave/contract-helpers": "npm:@ambiretest/aave-contract-helpers@1.3.1",
     "@aave/math-utils": "1.2.1",
     "@apollo/client": "^3.5.10",
     "@emotion/cache": "latest",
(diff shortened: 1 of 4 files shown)


defispartan:
Thanks for the contribution! I'll try to review later this week. 

Do you have a link to the batching EIP you mentioned so I can take a look at that as well?

0xdex18:
Sure!

Here is the batching EIP https://gist.github.com/mikhailxyz/773ba5196b5a7046423592204e5bc5c4

PR #792, opened 2022-05-19
feat: add coinbase browser

sakulstra:


MareskoY:
QC passed, but need additional verification on live

foodaka (approved):

PR #810, opened 2022-05-27
fix: fix dappbrowser autoconnect

sakulstra:


sendra (approved):

src/libs/web3-data-provider/Web3Provider.tsx
```diff
@@ -148,14 +148,16 @@ export const Web3ContextProvider: React.FC<{ children: ReactElement }> = ({ chil
     const { ethereum } = window;
 
     if (!ethereum?.providers) {
-      return false;
+      return true;
```
sakulstra, line 151: if there are no `providers` that means there is only one, so we should connect to it

src/libs/web3-data-provider/Web3Provider.tsx
```diff
@@ -148,14 +148,16 @@ export const Web3ContextProvider: React.FC<{ children: ReactElement }> = ({ chil
...
     }
 
     let provider;
     switch (providerName) {
       case 'CoinBase':
-        //@ts-expect-error no type
-        provider = ethereum.providers.find(({ isCoinbaseWallet }) => isCoinbaseWallet);
+        provider = ethereum.providers.find(
```
sakulstra, line 157: both are now valid cb flags

MareskoY:
QC Passed, auto connect working

PR #811, opened 2022-05-27
feat: add seatbelt report to governance

sakulstra:
Adds a link to the according seatbelt report to governance.

@iamanastasia plz have a look if that's okay or should be differently styled.
It's an external link similar to the discussion one, but linking to the seatbelt/simulation report.

image


iamanastasia:
@sakulstra Hey! 
Looks good to me. Although:
1. I think we don't have this 'blue outlined button' style anywhere yet. Can we stick to grey outlined button? 
2. Could the buttons be aligned to the left if they are put together?
3. Why is it 'simulation' report instead of 'seatbelt' report? I mean we can explain this in the UI if needed + provide some links to the governance. 

Here is a screenshot for alignment and button style
Screenshot 2022-05-30 at 13 05 55

PR #818, opened 2022-05-30
feat: seatbelt integration

sakulstra:
- supersedes #811

image

@iamanastasia plz have another look


.github/workflows/update-cache.yml
```diff
@@ -8,7 +8,7 @@ concurrency:
 on:
   workflow_dispatch:
   schedule:
-    - cron: 0 */12 * * *
+    - cron: 0 */3 * * *
```
sakulstra, line 11: the cache update won't do anything when there's no update so it's reasonable imo to run this more frequently so the update happens faster in case there is an update.

README.md
```diff
@@ -26,6 +26,7 @@ Enabling users to:
 Install it and run:
 
 ```sh
+cp .env.example .env.local
```
sakulstra, line 29: was missing, but needed when initially setting this up

defispartan (approved):

foodaka (dismissed):

defispartan (approved):

foodaka (approved):

PR #823, opened 2022-06-01
market configs

foodaka:


src/ui-config/marketsConfig.tsx
```diff
@@ -101,6 +102,26 @@ export const marketsData: {
...
+    marketTitle: 'Ethereum Ropsten',
+    chainId: ChainId.ropsten,
+    enabledFeatures: {
+      faucet: true,
+      governance: true,
+      staking: true,
+      incentives: true,
+    },
+    rpcOnly: true,
+    addresses: {
```
foodaka, line 115: @kartojal can u have a quick look and let me know if these look correct

kartojal:
would be cool to do some (manual) testing of all assets, but ipfs gateway is not loading for me

foodaka:
@bojanaave can you test all the assets on ropsten

bojanaave:
HI @foodaka  after performing smoke testing on the Ropsten network I did not spot any issue.

kartojal (approved):

PR #830, opened 2022-06-03
fixes for eth pricing

MareskoY:
eth pricing fixing for mainnet tests


foodaka (approved):

foodaka:
@MareskoY how does bumping these values fix the test?

PR #841, opened 2022-06-07
fix: only show testnet markets in staging environment

grothem:
Only show testnet markets in staging environment, and remove the testnet toggle switch in the setting menu. In prod, allow testnets to be toggled on, which is the current functionality.

- Removed the NEXT_PUBLIC_ENABLE_TESTNET env var, and replaced it with NEXT_PUBLIC_ENV, which can be either 'staging' or 'app'. This makes it more explicit which environment to use and when we should show/hide testnets. This also simplifies local development too because the value of the testnet toggle was saved to local storage. So to test a 'staging' environment, you'd have to update the env var, and also clear local storage.
- Only show the TestnetModeSwitcher if NEXT_PUBLIC_ENV is set to 'app'
- Only show testnet markets in the dropdown for 'staging' environment
- Show testnet markets and production markets if the testnet mode toggle is on


src/layouts/SettingsMenu.tsx
```diff
@@ -69,7 +69,7 @@ export function SettingsMenu() {
         </MenuItem>
 
         <DarkModeSwitcher component={MenuItem} />
-        <TestNetModeSwitcher component={MenuItem} />
+        {process.env.NEXT_PUBLIC_ENV === 'app' && <TestNetModeSwitcher />}
```
foodaka, line 72: hmm should we name this "app" or something like prod or production

grothem: Yeah that makes more sense, I'll update it

grothem: Changing it to prod

Issue #846, opened 2022-06-08
Deprecate Kovan Market


foodaka:
**Is your feature request related to a problem? Please describe.**
Kovan is being deprecated and removed from the test net markets

**Describe the solution you'd like**
Remove Kovan from v2 market and amm market

Issue #697, opened 2022-04-21
Make it clearer that staking is only availalbe on the ETH network


helpfulBro:
**Is your feature request related to a problem? Please describe.**
When a user visits the stake menu and is not on the ETH network the UI elements are disabled but there is no copy advising how they can stake.

**Describe the solution you'd like**
Include some copy and perhaps a CTA button allowing the user to switch to the Ethereum network to interact with the staking page.

**Describe alternatives you've considered**
None

**Additional context**
Image of disabled stake page when not on ETH network
![image](https://user-images.githubusercontent.com/25915621/164567808-9cf57098-3b22-44aa-a1fd-c6f3dc422171.png)

iamanastasia:
Quick questions/comments to define problem more accurately. 

1. Probably 'Stake' buttons are disabled because there is 0 AAVE/ABPT balance in the wallet. Here are my screenshots where there is AAVE balance and no ABPT balance. 
![step 1](https://user-images.githubusercontent.com/92739089/170474173-e49af013-c78e-40e6-9cca-0d9acace6a67.png)

**Solution suggestion #1:** Add AAVE/ABPT wallet balance and disable 'Stake' button when balance = 0. 
![Ethereum network](https://user-images.githubusercontent.com/92739089/170474407-0096d4c8-1e0f-4424-96dc-93ab7f5b015a.png)


2. Staking shows Mainnet info and suggests to switch Network only in tx modal window, which is not very friendly if you have 0 balance and can't open any tx window. (screenshot proof) 
![step 2](https://user-images.githubusercontent.com/92739089/170474258-cf3d912d-49e2-4f03-b3bc-be0e850ec064.png)

**Solution suggestion #2:** Add 1 more place to show current network and suggestion to switch it.
![Wrong network v1](https://user-images.githubusercontent.com/92739089/170474462-2fe11d7a-6d0d-4e45-afc3-a81f3d65336c.png)


3. There is Testnet issue. Whenever testnet switcher is enabled App shows Kovan info. And I see 2 problems here: a) It's not obvious you are in the testnet (where is my balance?!) b) no identification that it should be Kovan network. 

**Solution suggestion #3:** Add big non-dismissable banner on top across all sections whenever testnet switch is turned on. 
![Staking on Testnet](https://user-images.githubusercontent.com/92739089/170474539-8c832a2e-4940-45ae-a0de-a922a359152e.png)

[...]

PR #847, opened 2022-06-08
remove kovan market and kovan amm market

foodaka:
closes  https://github.com/aave/interface/issues/846

### DIFF
--- a/src/components/MarketSwitcher.tsx
+++ b/src/components/MarketSwitcher.tsx
@@ -33,7 +33,7 @@ export const getMarketInfoById = (marketId: CustomMarket) => {
 };
 
 const getMarketHelpData = (marketName: string) => {
-  const testChains = ['Kovan', 'Rinkeby', 'Mumbai', 'Fuji', 'Testnet'];
+  const testChains = ['Rinkeby', 'Mumbai', 'Fuji', 'Testnet'];
   const arrayName = marketName.split(' ');
   const testChainName = arrayName.filter((el) => testChains.indexOf(el) > -1);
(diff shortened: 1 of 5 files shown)


sakulstra (approved):

sakulstra (changes requested):
When removing the network config but keeping it in staking it will break.

I think either deploying a new testnet staking, or completely removing it would be reasonable.
In governance we only have mainnet as well and noone ever had a problem with that as it's more reasonable to test on a fork anyways.

foodaka:
@sakulstra @defispartan this is the state now


https://user-images.githubusercontent.com/8342048/172652996-785bc7d7-60c9-4d51-961b-9fc4dce619f6.mov

defispartan:
> @sakulstra @defispartan this is the state now
> Screen.Recording.2022-06-08.at.16.08.33.mov

Not a huge fan of this, when I see this warning my next question would be "How do I access this on an Ethereum fork?", and there's no way to do this from the interface, so this will just create more support questions than it answers. In my opinion this pr should just be removing the Kovan markets, and ensuring that staking is still functional in normal and testnet mode.

Then we can decide in #857 how to display staking/governance in testnet mode, if at all.

foodaka:
yeah i think we can separate the PRs per issue and not let this creep in. The staking was referenced from this issue https://github.com/aave/interface/issues/697

foodaka:
This PR now only in scope for removing kovan as per comment above

JoaquinBattilana:
Kovan was removed on https://github.com/aave/interface/pull/1004, closing this now

PR #857, opened 2022-06-10
refactor: testnet ux [skip cypress]

defispartan:
Several enhancements to the testnet UX:

- Making faucet links more accesible
- Displaying *only* testnet markets when testnet mode is enabled, or when accessing staging (staging.aave.com)

TO-DO:

- [x] Update gitbook FAQ with testnet and staging info - @grothem (changes are ready, just waiting on this pr to merge)
- [x] Check that CI issues is working with new env variables and default to prod ENV when starting app with no .env set - @grothem
- [x] Top Bar with button to disable - @grothem (@defispartan will get text and links)
- [x] Disable Staking and governance in staging env or testnet mode - we'll re-add in https://github.com/aave/interface/pull/858 - @grothem
- [x] Dark tooltips for faucet button - @defispartan
- [x] Add V2/V3 tab to market switcher - @defispartan

PR #858, opened 2022-06-10
Feature/staking states

foodaka:
WIP staking state warnings that the user is not connected to ETH main OR they are on a test net

closes https://github.com/aave/interface/issues/697


grothem:
The plan is to include these changes with PR #857 that has the other testnet UX updates. Can I merge these changes into the branch for that PR?

foodaka:
@defispartan i think it is ok from @grothem ? i think we can remove the section about ethereum fork? wdyt?

PR #870, opened 2022-06-14
fix: exit render loop

sakulstra:
using parameterless render functions like `  const ReserveIcon = () => {` results in a complete remount on every state dispatch (1x per second). It's completely unnecessary here.

- adds svg optimization to lint-staged
- fixes issue with mixed case symbol names

### DIFF
--- a/package.json
+++ b/package.json
@@ -132,7 +132,8 @@
     "*.js": "eslint --cache --fix",
     "*.jsx": "eslint --cache --fix",
     "*.ts": "eslint --cache --fix",
-    "*.tsx": "eslint --cache --fix"
+    "*.tsx": "eslint --cache --fix",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -132,7 +132,8 @@
     "*.js": "eslint --cache --fix",
     "*.jsx": "eslint --cache --fix",
     "*.ts": "eslint --cache --fix",
-    "*.tsx": "eslint --cache --fix"
+    "*.tsx": "eslint --cache --fix",
+    "*.svg": "npx svgo"
```
sakulstra, line 136: running svg optimization over newly committed svgs

defispartan (approved):
🙌

grothem (approved):

PR #874, opened 2022-06-14
fix: update convex token name

AustinGreen:


AustinGreen:
@miguelmtzinf @foodaka let's get this merged to follow foodaka's suggestion of using the right erc-20 token name

defispartan (approved):

PR #888, opened 2022-06-24
remove: kovan and rinkeby config

defispartan:
- Remove Kovan and Rinkeby config (networks being deprecated at end of month)
- TO-DO: Add V2 and V3 deployments on Goerli testnet


defispartan:
Re-added ETH Rinkeby since this network will be maintained for at least another year

MareskoY (approved):

PR #894, opened 2022-06-27
feat: always default to v3 when opening market switcher

defispartan:
- Always default to V3 when opening modal
- Fix console error for out-of-range select value


grothem (approved):

foodaka (approved):
Thanks for fixing this one

PR #896, opened 2022-06-28
feat/rwc updated design

defispartan:
- New design for repay w/ collateral modal
- New design for collateral swap modal
- Factor slippage amount into the maximum repayable debt for the selected collateral. Now, if the user presses the 'max' button on the debt asset when repaying, the amount inserted will take into account the slippage. This only happens when the balance of the selected asset to repay with (minus slippage) is less than the debt amount.
- Frozen assets are no longer available to select in the repay with collateral modal
- The slippage for the swap modal was ignoring the user selected value and was just hard coded to 1%. This will now use the selected value in the modal


MareskoY:
need to merge main into this pr to fix the tests

drewcook:
How does this relate to #1219? Should this PR be closed out in favor of it?

Issue #891, opened 2022-06-24
Showing incorrect colors in liquidation risk parameters modal


defispartan:
When a user's LTV is approaching the LiquidationThreshold the risk parameters modal should show this warning in red, not green:

![liquidationParametersModal](https://user-images.githubusercontent.com/26660211/175664867-fc284e81-d443-4456-b10e-de6b4a6ede3c.PNG)

PR #906, opened 2022-06-30
make warning message red on liq params

foodaka:
Closes https://github.com/aave/interface/issues/891


defispartan:
I'm not sure why the user circled that block of text in their screenshot. 

The issue is with the chart colors. It's showing a green line and a green circle in the Current LTV section even though the LTV is extremely close to the liquidation threshold. This should be a gradient shifting to red as they get closer to the liquidation threshold, same as the Health Factor section

grothem:
Closing this. Fixed with #1070

PR #907, opened 2022-06-30
fix: diff calculation

sakulstra:
The calculation for the required diff is simply broken right now.

`proposal/41` required 15.6M votes in the current ui to pass differential, which is obviously wrong

I think things got broken when the ui switched from showing % to showing absolute numbers. Please tripple check my code.


grothem:
I was looking through this trying to understand the differential calculation, and I'm a little lost 😅

I saw that the minimum diff values are set to 50 and 150 for short and long, but I'm not really sure what that means and how that factors into the calculation. If there's some documentation you could point me to, or maybe walk through an example w/ numbers, that would be awesome 🙂

sakulstra:
@grothem the diff value is a percent representation where `10000`, would be 100% so, for short it's 0.5% and long 1.5%.

The formular for when it's valid you can find here https://github.com/aave/governance-v2/blob/master/contracts/governance/ProposalValidator.sol#L191 

The formula in human readable words i think would be:
`positive% > negative% + VOTE_DIFFERENTIAL%`
So transformed should be `positive%-negative% > VOTE_DIFFERENTIAL%` which when substituted should be sth like `positiveVotes-negativeVotes > differentialVotes` ... so i guess it's just `diff > 80k | 240k` dependent on long vs short

defispartan (approved):
Took me a few tries to wrap my head around this, but I think this new version is correct.

In your comment on this thread though you gave the differentials as 0.5% and 1.5%, when they should be 5% and 15%. The long timelock executor has a threshold of 1500, and since  `ONE_HUNDRED_WITH_PRECISION` is 10000 on the contract this is a differential of 1500 / 10000 = 15%.

On Proposal 41, this would give a required differential of 16,000,000 * 0.15 = 2,400,000 which is what I'm getting

sakulstra:
@defispartan it's 0,5% and 15% actually - just double checked and it's 50 and 1500

src/modules/governance/utils/formatProposal.ts
```diff
@@ -24,7 +24,10 @@ export function formatProposal(proposal: Omit<Proposal, 'values'>) {
 
   const diff = new BigNumber(proposal.forVotes).minus(proposal.againstVotes);
   const voteSum = new BigNumber(proposal.forVotes).plus(proposal.againstVotes);
-  const requiredDiff = voteSum.multipliedBy(new BigNumber(proposal.minimumDiff).dividedBy(100));
+
+  const requiredDiff = new BigNumber(proposal.totalVotingSupply)
+    .multipliedBy(proposal.minimumDiff)
+    .dividedBy(10000);
```
foodaka, line 30: should we label the 10000 as a variable so its more clear or leave a comment on how we do the calculation?

defispartan: Yeah, it would be good to move that to a variable and add this [link](https://github.com/aave/governance-v2/blob/master/contracts/governance/ProposalValidator.sol#L191) as a comment so we remember the calculation

sakulstra: Feel free to propose a comment. I will accept it. Imo it doesn't make to much sense - perhaps having a comment above the function with a link to the smart contract but no explaining a single random value.

sakulstra: https://github.com/aave/interface/pull/907/files#diff-31d08d0f7bfd01ee2d2d0001e717dd0bbc87035fa4774e4eca2b189dcf1f61b1R6

defispartan (approved):

foodaka (approved):

PR #919, opened 2022-07-05
remove ability to embed as iframe across other sites

foodaka:
Removes ability to embed aave website as iframe on other websites to reduce clickjacking attemps


next.config.js
```diff
@@ -22,4 +22,17 @@ module.exports = withBundleAnalyzer({
   // assetPrefix: "./",
   trailingSlash: true,
   pageExtensions,
+  async headers() {
```
sakulstra, line 25: Pretty sure this does not work. No async header on a static app hosted on ipad.

PR #935, opened 2022-07-12
feat: add reserve factor [skip cypress]

defispartan:
Display reserve factor in borrow info section of reserve overview page.

reservefactor

CollectorTooltip


foodaka (approved):

src/modules/reserve-overview/ReserveConfiguration.tsx
```diff
@@ -401,6 +402,25 @@ export const ReserveConfiguration: React.FC<{ reserve: ComputedReserveData }> =
                     <ReserveSubheader value={reserve.borrowCapUSD} />
                   </PanelItem>
                 )}
+                <PanelItem
+                  title={
+                    <ReserveFactorTooltip
+                      text={<Trans>Reserve factor</Trans>}
```
drewcook, line 408: Just looking at this and `TextWithTooltip.tsx` as a potential optimization.

We could eliminate having to pass in `text={<Trans>....</Trans>}` by just wrapping the text prop with the `Trans` component and be more DRY.  Doing so would allow any of these tooltip components to just take in `text="some text"` and it eventually get wrapped at the base level.

We'd update [this line](https://github.com/aave/interface/blob/main/src/components/TextWithTooltip.tsx#L27) to be `{text && <Typography {...rest}><Trans>{text}</Trans></Typography>}`

Thoughts?

foodaka: Nice catch @drewcook - i think its a nice little optmization

drewcook: Great question.  This is something worthy to test - make the change and run the extractor to see if it is indeed able to pick up on the dynamic values post-injection into the JSX templates.  I wonder if also running the extractor against the source files vs dist files after `yarn build` would make any difference to this scenario as well, just a side thought.  Of course we'll want it to run against our source files for daily development, though.  Regardless, I think it's worth testing and a great call out.

MareskoY (dismissed):

src/modules/reserve-overview/ReserveConfiguration.tsx
defispartan: @drewcook Could you create a separate issue to look in to this?

drewcook: I have been keeping an eye on this during development.  It looks like it _does not_ pick up on dynamic values.  It will stringify them with the JSX `{}` and an arbitrary variable for the value within it, rather than being able to compile it.  I believe this is because the i18n is manually run via compile/build time rather than runtime.  Those dynamic bits will be filled in with the proper values at runtime, and so the language interpreter wouldn't be able to guess those.

Here are some examples where dynamic bits are being used within JSX tags:

[one](https://github.com/aave/interface/blob/main/src/locales/de/messages.po#L105)
[two](https://github.com/aave/interface/blob/main/src/locales/fr/messages.po#L171)
[three](https://github.com/aave/interface/blob/main/src/locales/es/messages.po#L339)

Now, that said, the translation is able to translate anything else. But in the case where we'd want to pass in a large string/sentence as a prop and have it translated, I'm not sure the `.po` files will pick up on that, and at runtime, it may still be english, since that's what we wrote it in in code.

This said, I'm happy to put these findings into an issue though and fully test it if we think it's worth our while.

drewcook: New issue representing this - https://github.com/aave/interface/issues/960

MareskoY (approved):

grothem (approved):

grothem (dismissed):

src/modules/reserve-overview/ReserveConfiguration.tsx
```diff
@@ -431,6 +432,21 @@ export const ReserveConfiguration: React.FC<{ reserve: ComputedReserveData }> =
                   </ParentSize>
                 </ChartContainer>
               )}
+              <Box
+                sx={{ display: 'inline-flex', alignItems: 'center', pt: '42px', pb: '12px' }}
```
drewcook, line 436: For some of these `px` values, we can use the built in MUI values, which map to 1=4px.  So `pb: '12px'` can become `pb: 3` and `'42px'` can become `10.5`.

src/modules/reserve-overview/ReserveFactorOverview.tsx
```diff
@@ -0,0 +1,83 @@
...
+        <Link
+          href={explorerLinkBuilder({
+            address: collectorContract,
+          })}
+          target="_blank"
+          sx={{ textDecoration: 'none' }}
+        >
+          <Box sx={{ display: 'flex', alignItems: 'center' }}>
+            <Typography variant="description" color="text.secondary">
+              Link
```
drewcook, line 64: Could this be better display text for the link?

src/components/infoTooltips/ReserveFactorTooltip.tsx
```diff
@@ -0,0 +1,26 @@
...
+  collectorLink?: string;
+}
+
+export const ReserveFactorTooltip = ({ collectorLink, ...rest }: ReserveFactorTooltipProps) => {
+  return (
+    <TextWithTooltip {...rest}>
+      <Trans>
+        Reserve factor is a percentage of interest which goes to a{' '}
+        {collectorLink ? (
+          <Link href={collectorLink} target="_blank">
```
drewcook, line 16: Please add `rel="noopener"` for any `target="_blank"` links as a best practice.  We'll want to practice this with new additions.  I recently added them to all existing new window links.  This attribute helps protect us because it prevents the website we link to from gaining access to our page/app via the window object.

drewcook (changes requested):
Just a few small things that could be improved, nothing major in the code.

src/modules/reserve-overview/ReserveConfiguration.tsx
defispartan: If this page is ever re-designed, it makes sense to refactor the styling and change these. For now I think it's fine to leave it though

src/modules/reserve-overview/ReserveFactorOverview.tsx
defispartan: I matched the color from the design.

I actually tested this with `text.primary` instead, and it looked much worse having the link the same color as the text/number fields

src/components/infoTooltips/ReserveFactorTooltip.tsx
defispartan: These are added automatically in our Link primitive: https://github.com/aave/interface/blob/92269df64a68a3502cb0fd8ff77f399e8503af0e/src/components/primitives/Link.tsx#L90, so my `target="_blank"` here is useless

drewcook: Right... I made that change, ha!

src/modules/reserve-overview/ReserveFactorOverview.tsx
drewcook: I was commenting mainly on the verbiage `Link`.  Maybe something like `View Contract` or whatever is most relevant and less generic?

src/components/infoTooltips/ReserveFactorTooltip.tsx
drewcook: Digging deeper on this topic, we can also support an `external` prop for `<Link/>`, as _we are assuming all uses of Link will be opening in a new tab_.  There's nothing wrong with that approach, but in cases where we don't want that to happen, we'd have the flexibility.  We could then add `target="_blank" rel="noopener"` only when the `external` prop is passed in.  Just some extra thoughts on this..

drewcook: Wow, disregard this last comment, should've read it first!

defispartan: The `Link` detects whether it's external based on the url (starting with 'http'). It's probably not as robust as an external prop, but good enough for most cases

src/modules/reserve-overview/ReserveFactorOverview.tsx
```diff
@@ -0,0 +1,83 @@
...
+          <Typography variant="description">
+            <Trans>Collector Contract</Trans>
+          </Typography>
+        }
+      >
+        <Link
+          href={explorerLinkBuilder({
+            address: collectorContract,
+          })}
+          target="_blank"
```
drewcook, line 59: Suppose these uses should be changed to `isExternal` now..

drewcook: Disregard this, but in response to your comment, yeah a prop may seem redundant when passing in a link elsewhere, but is more robust than the current check.

drewcook (changes requested):
The only action items after discussion seem to be just removing the redundant `target="_blank"` references on `Link`, since that's determined and added under the hood.

defispartan:
> The only action items after discussion seem to be just removing the redundant `target="_blank"` references on `Link`, since that's determined and added under the hood.

Sounds good, I'm updating the "Link" text to "View contract" as well

src/modules/reserve-overview/ReserveFactorOverview.tsx
drewcook: @defispartan Last ref here

defispartan: Fixed

drewcook (approved):
Looks good 👍

PR #943, opened 2022-07-17
improve-e-mode-menu-ux

vishesh0123:


vishesh0123:
Tasks Done 

- [ ] E-Mode Disabled" should not be a menu option when no eMode is active

- [ ]  Change the "E-Mode Disabled" text to "Disable E-Mode", and potentially separate switch and disable e-Mode menus

defispartan:
Thank you!

Merged into #953, along with some other small changes to the text and logic of the E-mode menu

PR #946, opened 2022-07-18
perf: parallelize & reduce calls on governance

sakulstra:
Fetches pending proposals in parallel instead of fetching them in sequence
Removes unnecessary `getBlockNumber` call

Ppl reported https://app.aave.com/governance/ being slow, especially when the ache is outdated - this improves the situation.


defispartan (approved):
⚡

MareskoY (approved):

PR #953, opened 2022-07-19
feat: improve emode menu ux

defispartan:
- Supersedes #943 
- Update text on all e-Mode menu items and warning states
- Better UX for markets with multiple e-modes like Avax V3: specific `enable`, `switch`, and `disable` buttons on dashboard and modal states
- Added LTV and Liquidation Threshold changes to tx modal


MareskoY (approved):

vishesh0123 (approved):

src/components/transactions/Emode/EmodeModalContent.tsx
```diff
@@ -117,34 +126,30 @@ export const EmodeModalContent = () => {
...
-              FAQ
-            </Button>{' '}
-            to learn more.
-          </Trans>
+          <Warning severity="error" sx={{ mt: 4 }}>
+            <Trans>
+              Enabling E-Mode requires all borrowed assets to belong to the same category. To enable
+              a new E-Mode, all borrow positions outside of this asset category must be closed. See
+              our{' '}
+              <Link
```
drewcook, line 143: If we use our `Link` primitive, target/rel become inherited.

src/components/transactions/Emode/EmodeModalContent.tsx
```diff
@@ -49,50 +49,59 @@ export const EmodeModalContent = () => {
...
+
+          emodeCategoriesArray.push(emodeParams);
+        }
+      });
+      const emodeCategories: Record<number, EmodeCategory> = {};
+      emodeCategoriesArray.forEach((category) => {
+        emodeCategories[category.id] = category;
+      });
+      // Default settings of the modal window, depending on the user's active emode and # of available categories
+      const selectedEmodeId =
```
drewcook, line 84: One (non-blocking) improvement I can see that would be helpful is to have an enum for use of the `user.userEmodeCategoryId`, so that statements like these and throughout the React component JSX more readable.

drewcook: Would have side-effects and improve other areas like [here](https://github.com/aave/interface/blob/553342f5def9fc5fec087bdd829a85c42dca55ac/src/components/transactions/Emode/EmodeNaming.ts#L1)

vishesh0123 (approved):

src/components/transactions/Emode/EmodeModalContent.tsx
defispartan: Currently, we export a helper function: https://github.com/aave/interface/blob/main/src/components/transactions/Emode/EmodeNaming.ts to get the name based on id since the same id might be named differently on different networks

bojanaave (approved):

vishesh0123 (approved):

defispartan:
There's a few small design changes to make, will be ready soon™️

MareskoY:
@defispartan  there is a bug 

turn on button available without selecting e-mode category (also I think that the StableCoin option should be selected by default)

Screenshot 2022-08-17 at 21 13 47

defispartan:
> @defispartan there is a bug
> 
> turn on button available without selecting e-mode category (also I think that the StableCoin option should be selected by default)
> 
> Screenshot 2022-08-17 at 21 13 47

@MareskoY Fixed in latest commit

bojanaave:
LGTM .

bojanaave (approved):

src/components/transactions/Emode/EmodeModalContent.tsx
```diff
@@ -109,138 +129,273 @@ export const EmodeModalContent = () => {
...
+            <Trans>Liquidation risk</Trans>
+          </Typography>
+          <Typography variant="caption">
+            <Trans>
+              This action will reduce your health factor. Please be mindful of the increased risk of
+              collateral liquidation.{' '}
+            </Trans>
+          </Typography>
+        </Warning>
+      )}
```
grothem, line 247: Looks like the liquidation risk warning will always be shown, even if the users health factor isn't going to decrease. Will that cause confusion?

MareskoY (dismissed):

src/modules/dashboard/DashboardEModeButton.tsx
```diff
@@ -40,14 +45,27 @@ export const DashboardEModeButton = ({
...
+      reserves.forEach((reserve: ComputedReserveData) => {
+        if (eModes.indexOf(reserve.eModeCategoryId) === -1) {
+          eModes.push(reserve.eModeCategoryId);
+        }
+      });
+      setEmodes(eModes);
+      setReservesLength(reserves.length);
+    }
+  }, [reservesLength, setReservesLength, reserves]);
+
```
grothem, line 60: Small optimization, but I think you could get rid of the useEffect and the state for reservesLength and eModes and just calculate the eModes on every render

defispartan: Removed this, the extra hook is not worth it since the component is re-rendered on reserves change anyways

src/components/transactions/Emode/EmodeModalContent.tsx
defispartan: Added logic to only display on HF decrease

defispartan:
@MareskoY @bojanaave Only piece of logic that changed is adding a check that the liquidation risk warning is only shown if the HF is decreasing. That warning should be the final thing to test :)

LiquidationRiskWarningEmode




LiquidationRiskWarningEmode

grothem (dismissed):

src/modules/dashboard/DashboardEModeButton.tsx
```diff
@@ -195,16 +206,46 @@ export const DashboardEModeButton = ({
...
-              onClick();
-              handleClose();
-            }}
-            data-cy={`${isEModeDisabled ? 'emode-enable' : 'emode-disable'}`}
-          >
-            {isEModeDisabled ? <Trans>Enable E-Mode</Trans> : <Trans>Disable E-Mode</Trans>}
-          </Button>
+          {isEModeDisabled ? (
+            <Button
+              sx={{ width: '100%' }}
```
drewcook, line 211: I believe the same effect can be accomplished with the `fullWidth` MUI prop on these `Button` components

defispartan: Yep, would be cleaner, will change

defispartan: Updated

grothem (approved):

src/modules/dashboard/DashboardEModeButton.tsx
```diff
@@ -82,23 +91,25 @@ export const DashboardEModeButton = ({
             alignItems: 'center',
             position: 'relative',
             zIndex: 1,
-            bgcolor: 'background.paper',
-            px: 1.5,
+            bgcolor: isEModeDisabled ? (open ? '#EAEBEF' : '#F7F7F9') : 'background.paper',
```
drewcook, line 94: For this, you could replace and rely on theme, which may be more readable:

`bgColor: isEModeDisabled ? (open ? theme.palette.divider : theme.palette.background.surface) : theme.palette.background.paper`

drewcook: With that said, I think we could possibly even add in a new background color for disabled: `theme.palette.background.disabled` that is the same light/dark values that the divider color uses

drewcook (approved):

grothem (approved):

PR #955, opened 2022-07-19
fix: don't break on md ipfs [skip cypress]

sakulstra:
In the 1inch proposal https://gateway.pinata.cloud/ipfs/QmPccptdeMjEAUbJLEXRw8s8dnWq6aVdigaqMUJGQJQBGV is referenced as ipfs instead of https://gateway.pinata.cloud/ipfs/QmeGHtzBASPj2tQyMZnBpiFghbZvawgx4kJ6HJMNUM4EY4 so the json() parser breaks as it's not a json.
This pr will clone the stream (as you can only read it once) and first try to parse it to json, if it's not successfull it will parse it as text.

Late night fix, but this fixes the ui

### DIFF
--- a/src/modules/governance/utils/getProposalMetadata.ts
+++ b/src/modules/governance/utils/getProposalMetadata.ts
@@ -26,16 +26,26 @@ export async function getProposalMetadata(
   if (!ipfsResponse.ok) {
     throw Error('Fetch not working');
   }
+  const clone = await ipfsResponse.clone();
+  try {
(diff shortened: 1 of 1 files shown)


defispartan (approved):

defispartan:
Without this, https://app.aave.com/governance/ is stuck loading proposals indefinitely since it can't parse the 1INCH proposal

defispartan:
Added [skip cypress] to speed up the time to update cache and deploy once merged

0x4Graham (approved):

PR #958, opened 2022-07-21
fix: remove weird stuff

sakulstra:
some weird unnecessary things i stumbled over when doing other stuff

### DIFF
--- a/src/ui-config/reservePatches.ts
+++ b/src/ui-config/reservePatches.ts
@@ -15,6 +15,7 @@ export const SYMBOL_MAP: { [key: string]: string } = {
   UNIWBTCUSDC: 'UNI_WBTC_USDC',
   UNIWBTCWETH: 'UNI_WBTC_WETH',
   UNIYFIWETH: 'UNI_YFI_WETH',
+  fUSDT: 'USDT',
 };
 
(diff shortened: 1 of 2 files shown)


src/ui-config/reservePatches.ts
```diff
@@ -100,15 +101,6 @@ export function fetchIconSymbolAndName({
       symbol,
     };
   }
-  //for fantom usdt => fUSDT
-  if (/fUSDT/.test(symbol)) {
```
sakulstra, line 104: i guess this was done because the .e and ^1 above. This is a different case though as one these networks multiple assets have the prefix and we were to lazy to list all of them. With fUSDT the symbol map is the perfect place.

drewcook: I've noticed that for Avalanche V3 that we're pulling the `USDT` symbol as `USDt`.  I think this is coming from the smart contracts, and it seems like it's in this same category of naming/capitalization with USDT.  Unrelated to this specific PR and changes, but has me thinking about while looking at this..

drewcook (approved):

foodaka (approved):

defispartan (approved):

PR #959, opened 2022-07-21
feat: add in supply/borrow/debt cap warnings, tooltips, and status components

drewcook:
# General Changes

- Adds in supply cap circular status when supply cap exists for an asset
- Adds in borrow cap circular status when borrow cap exists for an asset - however, there are currently no assets with borrow caps on any network
- Adds in a linear status for debt ceiling for any isolated asset
- Adds in tooltips and warning components for supply cap, borrow cap, debt ceiling
- Displays tooltips amongst dashboard list item components and new status components, and in reserve overview actions panel
- Also displays on supply modal, borrow modal, and collateral swap modal

## Notes

- Leverages a `getAssetCapUsage` utility "hook" so that this info can be used easily wherever a `reserve` is available in the UI.  This pattern could be expanded to create a global alert/tooltip system and display them easily across the UI, reducing repetition.


src/components/infoTooltips/BorrowCapTooltip.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { TextWithTooltip, TextWithTooltipProps } from '../TextWithTooltip';
+import { Link } from '../primitives/Link';
+
+type BorrowCapTooltipProps = TextWithTooltipProps & {
+  borrowCap: AssetCapData;
+};
+
+export const BorrowCapTooltip = ({ borrowCap, ...rest }: BorrowCapTooltipProps) => {
+  // Don't show a tooltip when less than 98% utilized
+  if (borrowCap.percentUsed < 98) return null;
```
defispartan, line 14: To switch this from a fixed 98% display cap to a user-based display, then it probably makes sense to check this in `getAssetCapUsage` and pass an enum prop to the tooltip either:

'none': No cap on the asset
'negligent': There is a cap but will not affect user
'partial': Remaining cap or ceiling will affect user max supply or borrow
'blocked': Remaining cap or ceiling is 0 or asset frozen

defispartan: Then it can be re-used for debt ceiling and frozen as well

src/modules/dashboard/lists/ListItemWrapper.tsx
```diff
@@ -46,8 +51,10 @@ export const ListItemWrapper = ({
         </Link>
         {frozen && <FrozenWarning symbol={symbol} />}
         {!frozen && symbol === 'AMPL' && <AMPLWarning />}
+        {supplyCap.determineTooltipDisplay({ supplyCap })}
```
defispartan, line 54: After the frozen logic is added, this should be moved to a separate component which will determine the priority or stacked display of warnings (AMPL can be added as well):

`<ReserveWarning user={...} reserve={...} />`

defispartan: Created issue #1001

bojanaave:
HI @drewcook 
QA team checked changes in your PR and here are the list of the thing that we found : 
1.  Supply and Borrow graphs are not visible for most of the assets on V3
Screen Shot 2022-08-05 at 13 44 47

 
2. We did not find   borrow cap circular status on V2 nor V3 
Screen Shot 2022-08-05 at 13 31 26

3. Also we are confused which of the info needs to visible and where so e.g 
Debth celling line graph is visible here USDT-Avalanche  (but for 90%+ of the assets is not visible) - 
Screen Shot 2022-08-05 at 13 51 08
 - So do we have any list where we can see which graphs/ cap circular statuses etc.. need to be visible and for which assets?

drewcook:
@bojanaave Good finds, these all seem to be accurate though.  There are zero assets with a borrow cap, thus, no borrow cap status will show.  I have also only been able to find a couple assets with a debt ceiling, one being USDt on Avalanche V3.  There aren't any caps on any asset in V2 for the most part.

The way that I've been testing for each scenario is by manually updating the percentages [here](https://github.com/aave/interface/pull/959/files#diff-4b5063ea31673c11bcfdf83f54629429c11719fb83d34fafe5a0410b1356a173R86)

bojanaave:
Hi drew , is it possible to underline Learn more inside info pop-ups ? 
Screen Shot 2022-08-08 at 13 33 53

src/hooks/getAssetCapUsage.tsx
```diff
@@ -0,0 +1,128 @@
...
+      determineTooltipDisplay: ({ borrowCap, useDefaultTooltip }) =>
+        borrowCap ? (
+          <BorrowCapTooltip borrowCap={borrowCap} useDefaultTooltip={useDefaultTooltip} />
+        ) : null,
+    },
+    debtCeiling: {
+      percentUsed: debtCeilingUsage,
+      isMaxed: debtCeilingReached,
+      // percentUsed: 99.994,
+      // isMaxed: true,
```
drewcook, line 113: These commented out lines are what I've been using to quickly test scenarios for given screens/modals/warnings/etc - since not all live production networks are going to yield every possible scenario.

src/components/infoTooltips/BorrowCapTooltip.tsx
```diff
@@ -0,0 +1,50 @@
...
+  useDefaultTooltip?: boolean;
+};
+
+export const BorrowCapTooltip = ({
+  borrowCap,
+  useDefaultTooltip = false,
+  ...rest
+}: BorrowCapTooltipProps) => {
+  // Don't show a tooltip when less than 98% utilized
+  // if (borrowCap.percentUsed < 98) return null;
```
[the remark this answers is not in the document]
drewcook: Yep, I left these in as a _reminder_.  This was something that came up - i.e. when do we show tooltips.  It was initially decided that we just won't show these tooltips if it's not in "warning zone" - i.e. > 98%.  But then I saw designs with the tooltip showing when the status is still in green zone.  So it makes sense that we decide this from a product POV.

I added in the additional else clause on line 31 to handle these cases, and added a comment in Figma about it [here](https://www.figma.com/file/p8a8k8ESSCWmqCbfmF12hU?node-id=242:1050378#242875756)

Ultimately, yes, we'll remove these lines, but we need to figure out answers for the above first.

src/hooks/getAssetCapUsage.tsx
```diff
@@ -0,0 +1,128 @@
...
+  /*
+    Supply Cap Data
+    % of totalLiquidity / supplyCap
+    Set as zero if either is zero
+  */
+  let supplyCapUsage: number = asset
+    ? valueToBigNumber(asset.totalLiquidity).dividedBy(asset.supplyCap).toNumber() * 100
+    : 0;
+  supplyCapUsage = supplyCapUsage === Infinity ? 0 : supplyCapUsage;
+  const supplyCapReached = supplyCapUsage !== Infinity && supplyCapUsage >= 99.99;
```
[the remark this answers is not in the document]
drewcook: I believe the `dividedBy` result can actually come back as `Infinity` anytime `asset.supplyCap` is zero in this case.  It would be useful to abstract the `x / 0` case within the utility function itself though so this logic is cleaner and it's centralized to that particular utility.

src/modules/dashboard/lists/SupplyAssetsList/SupplyAssetsList.tsx
```diff
@@ -89,6 +88,7 @@ export const SupplyAssetsList = () => {
         return [
           {
             ...reserve,
+            reserve,
```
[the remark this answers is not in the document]
drewcook: Right, so basically the values from `reserve` are depended on for finding out these caps and usage.  And these warnings/tooltips/status components can be deeply nested in a component hierarchy, very far from the original page components where this reserve data is available.  So rather than passing through the several reserve fields needed through as _multiple_ props all the way through to the nested component(s) that need them to show the warning, its cleaner to pass through just the single reserve.  Now, for the dashboard lists, most of the reserve fields are passed through as individual props when they're spread with `{...item}`.  However, not _all_ fields were passed in, and rather than ensuring multiple fields were plumbed through correctly, and/or creating a new representation in TypeScript, it was a path of least resistance to pass in `reserve` is it's own singular prop and allow the utility function to just work with a fully represented `ComputedReserveData` type.

This isn't ideal.  This is redundant, since most _but not all_ reserve fields are passed through to the components, but not all fields are continuously passed through.  For example, `SupplyAssetsListItem` may have access to all the reserve fields, but `ListItemWrapper` doesn't, as it's more general and has a more restrictive props list.

I think the next step to reduce this plumbing and redundancy of passing through the reserve entity would be to create a React context to hold this info and wrap the page level with it.  Then the deeply nested component can just access the context rather than a plumbed through prop.

drewcook:
@bojanaave I've added in underlining the learn more links

src/components/ReserveOverviewBox.tsx
```diff
@@ -27,9 +32,11 @@ export function ReserveOverviewBox({ title, children }: ReserveOverviewBoxProps)
...
       >
-        <Typography variant="secondary14" color="text.secondary" component="span">
-          {title}
-        </Typography>
+        {title && (
+          <Typography variant="secondary14" color="text.secondary" component="span">
+            {title}
+          </Typography>
+        )}
         {children}
```
drewcook, line 40: These changes just allow the component to be used full width rather than a 1/3, and an optional title, as the debt ceiling box doesn't use one.

src/components/TextWithTooltip.tsx
```diff
@@ -43,9 +45,9 @@ export const TextWithTooltip = ({
...
             sx={{
               fontSize: iconSize,
-              color: open ? 'info.main' : 'text.muted',
+              color: color ? color : open ? 'info.main' : 'text.muted',
               borderRadius: '50%',
-              '&:hover': { color: 'info.main' },
+              '&:hover': { color: color || 'info.main' },
             }}
           >
             {icon || <InformationCircleIcon />}
```
drewcook, line 53: Allows the styling of color on the tooltip icon, i.e. for red/yellow exclamation tooltips.

src/hooks/getAssetCapUsage.tsx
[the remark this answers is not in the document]
grothem: Yeah I had the same thought about creating utility function for that.

The result coming back as `Infinity` makes sense, but on line 56 you handle that case. So on line 57, `suppleCapUsage !== Infinity` should always be true.

bojanaave:
Looks fine. 
NOTE: Screenshots below are with fake values (that has been used just for testing purpose)

Screen Shot 2022-08-09 at 17 20 07
Screen Shot 2022-08-09 at 17 20 37
Screen Shot 2022-08-09 at 17 23 05

bojanaave (approved):

src/hooks/useAssetCaps.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+export type AssetCapData = {
+  percentUsed: number;
+  isMaxed: boolean;
+};
+
+export type AssetCapHookData = AssetCapData & {
+  determineWarningDisplay: (props: WarningDisplayProps) => JSX.Element | null;
+  determineTooltipDisplay: (props: WarningDisplayProps) => JSX.Element | null;
+};
```
drewcook, line 27: Some additional fields I think that could help DRY things up relating to these caps are:

- Status color to be used for styling the colors - based off percentUsed
- Tooltip verbiage for good/warning/max
- Warning verbiage for good/warning/max

I think we continue to allocate the logic for displaying these warnings/tooltips to the component that is implementing them.  But we could clean up some of repetitiveness of the tooltip, warning, and status components themselves by abstracting the common stuff to this context.  If you notice, the verbiage is nearly identical, the colors are, and this could be another improvement to this pattern.

```js
type AssetCapWarningTooltipVerbiage = {
   safe: string;
   warning: string;
   maxed: string;
 }
 
type AssetCapStatusColor = "success.main" | "warning.main" | "error.main"

export type AssetCapHookData = AssetCapData & {
  determineWarningDisplay: (props: WarningDisplayProps) => JSX.Element | null;
  determineTooltipDisplay: (props: WarningDisplayProps) => JSX.Element | null;
  tooltipVerbiage: AssetCapWarningTooltipVerbiage;
  warningVerbiage: AssetCapWarningTooltipVerbiage;
  statusColor: AssetCapStatusColor;
```

Then the logic that is in the tooltip components and colors on the status components becomes much more simpler, and maybe the tooltip/warning components become a singular component.

Thoughts on this addition?

bojanaave:
LGTM.
Screen Shot 2022-08-18 at 20 02 29
Screen Shot 2022-08-18 at 20 15 14

bojanaave (approved):

src/components/infoTooltips/BorrowCapMaxedTooltip.tsx
```diff
@@ -0,0 +1,30 @@
...
+};
+
+export const BorrowCapMaxedTooltip = ({ borrowCap, ...rest }: BorrowCapMaxedTooltipProps) => {
+  if (!borrowCap || !borrowCap.isMaxed) return null;
+
+  return (
+    <Box sx={{ ml: 2 }}>
+      <TextWithTooltip {...rest} icon={<ExclamationIcon />} color="warning.main" iconSize={18}>
+        <>
+          <Trans>Protocol borrow cap at 100% for this asset. Further supply unavailable.</Trans>{' '}
```
defispartan, line 19:
```suggestion
          <Trans>Protocol borrow cap at 100% for this asset. Further borrowing unavailable.</Trans>{' '}
```

defispartan (approved):
Big overhaul of warning states, much needed!

 Tested cap display on all affected screens with different market conditions, LGTM 👍

bojanaave:
Looks fine. 
Screen Shot 2022-08-22 at 20 51 05
Screen Shot 2022-08-22 at 20 53 46

bojanaave (approved):

PR #961, opened 2022-07-22
fix: switch default env from staging to prod

defispartan:
When running the app locally and following the steps from README, the app loads in `staging` env, which is confusing since it disables governance, staking, and production markets


drewcook:
Good update.

Maybe we also update `CONTRIBUTING.md` line 10 to copy to `env.local` instead of `.env`, unless that's an important difference?  Also, line 32 of the same file, should we keep as is?

drewcook:
Yeah, we may want to change it on line 32 as well since `prod` will be copied over.  Maybe we add a comment in the `.env.example` that says `# Update this to 'staging' to enable testnet mode only, which will eliminate governance, staking, and production markets`

defispartan:
> 

`.env.local` works for both development and production, and has higher precedence than `.env` https://www.querythreads.com/what-is-the-difference-between-env-local-and-env-development-local/ 

Definitely makes sense to update CONTRIBUTING, I'll do that now

drewcook (approved):
LGTM 👍

grothem (approved):

PR #962, opened 2022-07-22
fix: governance cache failure

defispartan:
Build cache action is failing due to #992. 

Voting data is not currently rendered on the UI, so it can be disabled for now while the root cause is investigated.


grothem (approved):

drewcook (dismissed):
LGTM

sakulstra:
I don't understand why this pr exists and why it was approved.

As i understand it there was an issue on the rpc returning ancient broken blocks.
This "fix", as i understand it simply ignores that issue allows storing "broken events" in the cache for eternity.

defispartan:
> As i understand it there was an issue on the rpc returning ancient broken blocks. This "fix", as i understand it simply ignores that issue allows storing "broken events" in the cache for eternity (broken data will never be re-cached).
> 
> The "bug" it tries to resolve resolved itself when the rpc recovered. Would recommend to not merge this.

Agreed, the cache error seems to have resolved itself so it makes no sense to merge this

defispartan:
Won't close #992, but is a temporary solution to fix the build cache

sakulstra (approved):

MareskoY (approved):

PR #964, opened 2022-07-25
feat: zustand / immer

sakulstra:
This pr replaces most part of the global state storage in `react-context` & `apollo-client` with `zustand`.
The advantages of this are alignment of data storage which allows better debuggability with e.g. `redux dev tools` as most of the state is stored in zustand.

This pr also removed the caching layer which is only used on two legacy pools right now.

Other positive side effects:
- Zustand also is framework agnostic so potentially one could get rid of react without having to rewrite all the application logic.
- Smaller bundle size (>40kb accross all pages)
- Less centralization (no more reliance on a 3th party api)
- better separation of ui and transaction logic


src/store/poolSlice.ts
```diff
@@ -0,0 +1,434 @@
...
+        user,
+      });
+    },
+    claimRewards: async ({ selectedReward }) => {
+      // TODO: think about moving timestamp from hook to EventEmitter
+      const timestamp = dayjs().unix();
+      const reserves = selectFormattedReserves(get(), timestamp);
+      const currentAccount = get().account;
+
+      const allReserves: string[] = [];
```
sakulstra, line 384: I'm pretty confident this is wrong.
It's straight from the ui so it's wrong on the interface already but might make sense to fix.

The point is it doesn't really make sense to pass all the assets as we already know which assets have rewards and which don't via `userIncentiveData` but instead we pass every single token of the market (which has incentives).
Am i misreading this?

defispartan: This is correct, there was a change made to the V3 IncentivesController to not store the accrued rewards in a separate variable after withdraw/repay, so passing the entire array to `assets` is the only way to ensure that rewards can be claimed for tokens that are no longer supplied or borrowed. https://github.com/aave/interface/pull/704

src/components/transactions/Repay/CollateralRepayActions.tsx
```diff
@@ -44,35 +40,26 @@ export const CollateralRepayActions = ({
   maxSlippage,
   ...props
 }: RepayActionProps) => {
-  const { lendingPool } = useTxBuilderContext();
-  const { currentChainId: chainId, currentNetworkConfig } = useProtocolDataContext();
-  const { currentAccount } = useWeb3Context();
+  // do we need currentAccount in deps?
+  const currentAccount = useRootStore((state) => state.account);
```
defispartan, line 44: Other methods like `supply` are accessing `currentAccount` inside of `poolSlice` so this isn't needed here

src/components/transactions/Faucet/FaucetActions.tsx
```diff
@@ -13,13 +13,13 @@ export type FaucetActionsProps = {
 };
 
 export const FaucetActions = ({ poolReserve, isWrongNetwork, blocked }: FaucetActionsProps) => {
-  const { faucetService } = useTxBuilderContext();
+  const mint = useRootStore((state) => state.mint);
   const { currentAccount } = useWeb3Context();
```
defispartan, line 17: I noticed in a few places `useWeb3Context` is used to get the `currentAccount`, should this be refactored to use the root store?

defispartan, line 17: I noticed in a few places useWeb3Context is still used to get the currentAccount, should this be refactored to use the root store?

src/hooks/governance-data-provider/useVotingPower.tsx
```diff
@@ -1,45 +1,7 @@
...
-      stkAaveVotingDelegatee @client
-      stkAavePropositionDelegatee @client
-      aaveTokenPower @client
-      stkAaveTokenPower @client
-    }
-  }
-`;
+import { useRootStore } from 'src/store/root';
 
 export const useVotingPower = () => {
```
defispartan, line 3: This entire hook could just be removed now lol

src/components/transactions/Faucet/FaucetActions.tsx
sakulstra: This pr touches 91 files and there have been quite massive merge conflicts down the road, so at some point we decided to "touch as little as possible".

With the currentAccount as long as aave-interface relies on outdated web3-react there's not much advantage from one over the other. The currentAccount in zustand is just the synced currentAccount from context.
Might make sense to refactor down the road, but for now i'd keep it (same reason for still having appDataProvider).

src/components/transactions/Repay/CollateralRepayActions.tsx
sakulstra: good catch, @satanworker can you please have a look at all the actions? I think most of them fetch & pass the account which should be handled inside the action.

src/components/transactions/UnStake/UnStakeActions.tsx
```diff
@@ -25,12 +25,12 @@ export const UnStakeActions = ({
...
 }: UnStakeActionProps) => {
   const { currentAccount } = useWeb3Context();
-  const stakingService = useStakeTxBuilderContext(selectedToken);
+  const redeem = useRootStore((state) => state.redeem);
 
   const { action, loadingTxns, mainTxState, requiresApproval } = useTransactionHandler({
     tryPermit: false,
     handleGetTxns: async () => {
-      return stakingService.redeem(currentAccount, amountToUnStake.toString());
+      return redeem(selectedToken)(currentAccount, amountToUnStake.toString());
```
sakulstra, line 33: @satanworker e.g. here the account is kinda unnecessary

src/components/transactions/Withdraw/WithdrawActions.tsx
```diff
@@ -27,31 +25,19 @@ export const WithdrawActions = ({
   blocked,
   sx,
 }: WithdrawActionsProps) => {
-  const { lendingPool } = useTxBuilderContext();
-  const { currentChainId: chainId, currentMarketData } = useProtocolDataContext();
   const { currentAccount } = useWeb3Context();
```
sakulstra, line 28: or here

src/hooks/governance-data-provider/useVotingPower.tsx
sakulstra: some of these hooks are essentially selectors now and don't really make sense.
That said we kept them as is to not increase the amount of files changed /merge conflicts.
We'll do a follow up pr for sure prettifying some of these things.

src/components/transactions/Repay/CollateralRepayActions.tsx
satanworker: done

src/components/transactions/UnStake/UnStakeActions.tsx
satanworker: yep, removed

src/components/transactions/Withdraw/WithdrawActions.tsx
satanworker: Moved current account from Actions to zustand,

MareskoY:
QC passed 

But required code review

satanworker:
@defispartan could you give it another go, please? I think all the issues fixed

src/store/poolSlice.ts
```diff
@@ -0,0 +1,444 @@
...
+      try {
+        promises.push(
+          poolDataProviderContract
+            .getReservesHumanized({
+              lendingPoolAddressProvider,
+            })
+            .then((reservesResponse) =>
+              set((state) =>
+                produce(state, (draft) => {
+                  if (!draft.data.get(currentChainId)) draft.data.set(currentChainId, new Map());
```
grothem, line 142: I think it'd be good to fix the non-null assertions, so we're not bypassing our `strict` compiler options.

I noticed our linter config doesn't pick this up as warning, it's an issue with next js lint directory config. We'll fix that separately, but if you run `next lint --dir src` you should see those warnings get flagged.

src/store/poolSlice.ts
```diff
@@ -0,0 +1,444 @@
...
+              .getUserReservesHumanized({
+                lendingPoolAddressProvider,
+                user: account,
+              })
+              .then((userReservesResponse) =>
+                set((state) =>
+                  produce(state, (draft) => {
+                    if (!draft.data.get(currentChainId)) draft.data.set(currentChainId, new Map());
+                    if (!draft.data.get(currentChainId)?.get(lendingPoolAddressProvider)) {
+                      draft.data.get(currentChainId)!.set(lendingPoolAddressProvider, {
```
grothem, line 172: same comment as above with the non-null assertions.

grothem (commented):
Looking really good overall!

I did have a question about some of the things that are being stored on the `PoolSlice` state. For example, `borrow`. It's called from the `BorrowActions` component, and is basically just a helper function to create the transaction. I'm wondering if it makes sense to hoist this kind of logic into the global state, since it's only used in 1 component. I noticed that most of the actions components have a similar pattern.

satanworker:
> Looking really good overall!
> 
> I did have a question about some of the things that are being stored on the `PoolSlice` state. For example, `borrow`. It's called from the `BorrowActions` component, and is basically just a helper function to create the transaction. I'm wondering if it makes sense to hoist this kind of logic into the global state, since it's only used in 1 component. I noticed that most of the actions components have a similar pattern.

I think it's totally ok to store methods in store if we are not 100% sure this actions won't be used anywhere else. Because in store, we have all the data available at once. 
About Actions component, I think that we would need to get rid of it in the next PR. Having transaction logic in context makes no sense 
take a look at BGD example here 
https://github.com/bgd-labs/fe-shared-examples/blob/acc9aa42e6bb9d934ee9de226ce5b27cd08b60b2/front-end/src/counter/store/counterSlice.ts#L28  
`executeTx` is doing all the logic of switching the network before submitting tx, saving transaction receipt to local storage, waiting for tx to execute etc
I think we'll need to build something similar and encourage calling transactions from inside the store and offload handling transactions to a single place, but I'm open to suggestions of course

src/store/protocolDataSlice.ts
```diff
@@ -0,0 +1,51 @@
...
+import { setQueryParameter } from './utils/queryParams';
+
+export interface ProtocolDataSlice {
+  currentMarket: CustomMarket;
+  currentMarketData: MarketDataType;
+  currentChainId: number;
+  currentNetworkConfig: NetworkConfig;
+  jsonRpcProvider: () => providers.Provider;
+  setCurrentMarket: (market: CustomMarket) => void;
+}
```
drewcook, line 22: `ProtocolData` seems too generic of a name for this.  This could be improved in one of two ways:

1. Rename to `MarketNetworkSlice`
2. Split out to two slices `MarketSlice` and `NetworkSlice` (I prefer this over 1)

drewcook: Even though markets and networks have lots in common, they are two different things and could be separated.  Network config, jsonRpcProvider, and chain ID make sense to be in network slice, and the rest in the market slice.

src/store/root.ts
```diff
@@ -0,0 +1,68 @@
...
+export const useRootStore = create<RootStore>()(
+  devtools((...args) => {
+    return {
+      ...createStakeSlice(...args),
+      ...createProtocolDataSlice(...args),
+      ...createWalletSlice(...args),
+      ...createPoolSlice(...args),
+      ...createIncentiveSlice(...args),
+      ...createGovernanceSlice(...args),
+    };
```
drewcook, line 33: My feedback on this large PR is one thing - these slices should belong to a top-level key.  Similar to how a large global redux store scoped actions and reducers within a given slice to a top-level key. There are a couple reasons that this is useful:

1. When creating new methods and/or state values within any given slice, a developer now needs to know if there is a key within that slice exists in some other slice across some other file within the codebase, which creates opportunity for human-error and can create more time debugging.  Now if/when I want to add a new value/method to a given slice, I now have to check across other slice files to ensure that that name doesn't exist already, which is cognitive overload.  It could also cause side-effects - i.e. what happens when there are two keys with the same name within the global store, does one get overwritten?  This is problematic, and scoping the slices to a top-level key prevents this.  Common keys that I see this are ones like `address`, `account`, `balance`, etc.
2. When using `useRootStore` at the component level, now it becomes very clear as to where that value/method belongs to.  It can help with debugging, it can help to understand that X method is part of Y feature/part of the protocol.

I'd like to see something like this implemented:

```js
// alphabetized (ocd)
return {
  governance: ...createGovernanceSlice(...args),
  incentive: ...createIncentiveSlice(...args),
  market: ...createMarketSlice(...args), // see previous comments
  network: ...createNetworkSlice(...args), // see previous comments
  pool: ...createPoolSlice(...args),
  staking: ...createStakeSlice(...args),
  wallet: ...createWalletSlice(...args),
};
 ```
 
 This will require the changes across each `useRootStore` use, but this is a good change.   It becomes clear on what part of the store you're working with.
 
 ```js
 // this reads nicely
 const governance = useRootStore(state => state.governance)
 governance.submitVote({...})
 
 // or, syntax sugar
 const { submitVote } = useRootStore(state => state.governance)
 submitVote({...})
 ```

drewcook (changes requested):

src/store/root.ts
satanworker: Awesome feedback, I think it's a good idea, but just for a sake of different perspective, if you have 2 methods with the same name, but different parameters type check will not pass. 
Splitting code to different reducers was the hardest part of Redux IMO, and I love that in zustand it's way simpler to access different parts of store, so you don't need to think too hard in terms of context of each reducer/slice but rather always have a global store at your disposal at any given moment. 
It's still way easier to maintain than using context where everything is coupled to hooks and local state, but at the same time very simple abstraction layer without too many specifics. But I think having keyed slices is worth it anyway to avoid name collision of methods with the same signature

src/store/protocolDataSlice.ts
satanworker: Good one, in other projects I've had Web3Slice (I know it's even more generic name) which stored everything related to wallet, RPCs and services wrappers around typechain 
https://github.com/bgd-labs/fe-shared-examples/blob/main/front-end/src/web3/store/web3Slice.ts 
so it's more like a dependency injection slice which is responsible for providing proper instances of anything related to web3 to other slices. So we can just use `get().activeWallet` or `get().dataService.getSomeData()` being sure that everything is connected to proper RPC and signer. 
What do you think? I think it's similar to `NetworkSlice` but I don't think having only network related data would be that useful, so maybe we should introduce `Web3Slice`

sakulstra: I kept this name when starting off to be at least breaking as possible with your current code & naming.
`protocolDataSlice` contains only the things previously contained in `protocolDataContext`

src/store/root.ts
sakulstra: Personally i think this is a rather bad idea and is counter to the core ideas of zustand (being stupid simple, easy to debug global store).

Due to type safety all non primitive type overlaps would result in conflicts that would result in build time errors. Unnecessary nesting makes state management & diffing more complex and imo therefore is no benefit, but if it gets this pr merged would change it :shrug:

src/store/protocolDataSlice.ts
drewcook: Good points.  Because the two are so intertwined together and used with each other, I'm onboard with the idea of keeping it as a single slice that may represent all the very common pieces of data that we get from a web3 provider.  So yes, we can just rename this since the previous name was just way too generic, even if it's used in other apps or was used previously.  I think `Web3Slice` works, or maybe better for readability as `Web3DataSlice`, so then it reads as `web3Data.someMethodOrValue`.

src/store/root.ts
drewcook: If its going to cause too much conflict in debugging down the road, then I see value in _not adding_ it in, but I don't have the experience yet debugging zustand.  I wasn't aware of the type safety checking that would happen if two methods with the same signature are part of the global store.  I'm a big fan of keeping things simple stupid and clean, but also like the idea of the scoping in the context of the primitives, and even though at this point the store may be simple and there aren't any overlaps, there are unknowns as to any issues that could arise if more things are added in to assist with this UI.  However, I supposed though if it becomes an issue down the line then we can always revisit it and key them and make the necessary updates across the board.  After our talk today and learning about how a store has access to other parts of inherited stores then I'm okay with the flattened idea, but again, I don't know what it's like debugging this state with and without the top level keys.

src/store/protocolDataSlice.ts
satanworker: Looking at comments fresh after the weekend, I think it would be better to have the same structure as context, but move logic to `zustand`, at least for the first PR, so it would be easier to track how things were done before and how to do it with zustand, changing both implementation and the logic behind the implementation would make very frustrating to reason is the change was done because of zustand or because the logical error. 
If we can agree that current PR has no new bugs, and we understand the purpose of introducing zustand, I think it's better to merge and later introduce smaller PRs to focus on each particular feature, i.e. renaming, transactions flow, removing context etc.
The purpose of this PR is again to lay the groundwork for new features and unlock refactoring without huge merge conflicts

defispartan (approved):
Major step forward for improving state management in the app 🐻 

Captured the open feedback here to address in future PRs: https://github.com/aave/interface/discussions/1293

src/store/protocolDataSlice.ts
foodaka: Agree with @satanworker for now lets first handle the large re architecture and then we can move into refactoring some of the naming so we dont do too much as this PR is already quite large

foodaka:
This is a massive refactor improving the state management of the application. Well done

drewcook (approved):
Great to see the bundle size significantly smaller and the logic abstracted out away from the components as a high-level vision.  The first step in a longer process to really clean up this front-end.  Nice work!

PR #965, opened 2022-07-25
fix: fix fork setup

sakulstra:
When you are on a fork, the ui will NOT tell you to switch network when mm selects mainnet, which imo is a quite serious issue.

I think introduced here https://github.com/aave/interface/pull/887/files or even present before.

### DIFF
--- a/src/components/transactions/GovDelegation/GovDelegationModalContent.tsx
+++ b/src/components/transactions/GovDelegation/GovDelegationModalContent.tsx
@@ -7,6 +7,7 @@ import { useState } from 'react';
 import { DelegationType } from 'src/helpers/types';
 import { useAaveTokensProviderContext } from 'src/hooks/governance-data-provider/AaveTokensDataProvider';
 import { useModalContext } from 'src/hooks/useModal';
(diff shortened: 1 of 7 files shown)


grothem:
Good catch, yeah it was introduced in that PR because you couldn't stake on a fork before those changes. I think the same issue will be in the `StakeCooldownModalContent` and `StakeRewardClaimModalContent`, introduced here: https://github.com/aave/interface/pull/900

sakulstra:
@grothem pushed a fix for the others as well

grothem (approved):

sakulstra:
@grothem rebased and added https://github.com/aave/interface/pull/965/commits/ee6f64b77da804dc4ebfcbcda20cc37e85baf4a8 since you reviewed - this commit makes the ui show "<network> fork" on the switch network dialog and also in the wallet dropdown
image

bojanaave:
checked on fix/fix-fork-setup and looks fine

grothem:
Nice! Looks good

grothem (approved):

PR #968, opened 2022-07-25
feat: arbitrum goerli testnet

defispartan:
- Add Arbitrum Görli network and market config
- Replace Goerli market names with Görli


grothem (approved):

sakulstra:
Doesn't it make sense to remove optimism_kovan and arbitrum_rinkeby in the same run as the rpcs are o longer reliable for these networks?

defispartan:
> Doesn't it make sense to remove optimism_kovan and arbitrum_rinkeby in the same run as the rpcs are o longer reliable for these networks?

These RPCs are still reliable, I would assume Optimism/Arbitrum probably continue to run their own nodes on Kovan/Rinkeby to maintain them. 

Eventually, yeah it makes sense to remove, but neither has fully migrated yet. They don't have full block explorer or bridge support on Optimism Görli yet.

PR #973, opened 2022-07-26
feat: transak on-ramp integration

debajitr:


src/modules/staking/GetABPTokenModal.tsx
```diff
@@ -1,15 +1,15 @@
...
-import { Trans } from '@lingui/macro';
 
-type GetAPBTokenModalProps = {
+type GetABPTokenModalProps = {
   open: boolean;
   close: () => void;
 };
 
-export const GetAPBTokenModal = ({ open, close }: GetAPBTokenModalProps) => (
+export const GetABPTokenModal = ({ open, close }: GetABPTokenModalProps) => (
```
defispartan, line 12: Good catch 👍

src/ui-config/menu-items/index.tsx
```diff
@@ -83,6 +83,11 @@ const moreMenuItems: MoreMenuItem[] = [
     title: t`Switch to Aave Classic`,
     icon: <AaveClassic />,
   },
+  {
+    link: 'https://global.transak.com',
```
defispartan, line 87: If there are multiple `onRampService` providers this will need to change to display a modal or sublinks.

Since this will require additional design, and is complicated due to the component layout, it's fine to leave this for a future PR. Could you open an issue on the main interface repo for this task: 'Change menu links to support multiple on ramps'

defispartan (commented):
Looks really good!

To get this setup to test:

1.) Rebase or merge (whichever you prefer) the latest interface version into this PR. There are a few changes to .env variable setup since this was created.

2.) You can add `NEXT_PUBLIC_TRANSAK_URL` and `NEXT_PUBLIC_TRANSAK_API_KEY` to the file `env.local` and import them directly into `onRampServicesConfig` as you currently have setup.

src/modules/staking/BuyWithFiatModal.tsx
```diff
@@ -0,0 +1,66 @@
...
+  return (
+    <BasicModal open={open} setOpen={close}>
+      <Typography variant="h2">
+        <Trans>Buy Crypto with Fiat</Trans>
+      </Typography>
+      <Typography sx={{ my: 6 }}>
+        {onRampServices.length && onRampServices.length === 1 ? (
+          <Trans>
+            {onRampServices[0].name} on-ramp service is provided by External Provider and by
+            selecting you agree to Terms Terms of the Provider. Your access to the service might be
```
drewcook, line 30: "Terms Terms"

src/ui-config/menu-items/index.tsx
debajitr: Done https://github.com/aave/interface/issues/1031

src/modules/staking/BuyWithFiatModal.tsx
debajitr: Done

MareskoY:
there are 3 points over here

**Point1**

Buy Crypto With Fiat from header every time redirect to transak on Ethereum network with aave token

could we open transak interface with similar network as was a market in the interface (if it was an avalanche market) it should be open transak on the avalanche network 

with base asset by default

**Point2**

there are some count of assets which is not supported in transak but listed in aave like DAI on optimism 

in this case, transak open on the optimism network but the asset is ETH

https://global.transak.com/?apiKey=799087ea-c2be-44b8-a0fe-56d96b37ee3c&network=Optimism&cryptoCurrencyCode=DAI&walletAddress=0xe4217040c894e8873ee19d675b6d0eec992c2c0d&disableWalletAddressForm=true

most of the assets on the Optimism network


**Point3**
need to update pr with the current `main` branch, to see the result of auto testing  cc @debajitr

debajitr:
@MareskoY point 3 done

debajitr:
@MareskoY code changes for point 2 has been pushed, we will hide the button if it is not supported on our side. The API for the same will go live on 27th EOD IST.

debajitr:
@MareskoY point 1 is also handled

debajitr:
@defispartan @drewcook please re-review the changes and let me know if anything needs to be changed.

src/layouts/MoreMenu.tsx
```diff
@@ -56,7 +60,15 @@ export function MoreMenu() {
         keepMounted={true}
       >
         {moreNavigation.map((item, index) => (
-          <MenuItem component={Link} href={item.link} key={index}>
+          <MenuItem
+            component={Link}
+            href={
+              item.makeLink
+                ? item.makeLink({ cryptoSymbol: 'AAVE', network, walletAddress })
```
defispartan, line 67: Could you possibly remove the `cryptoSymbol` and `network` for this specific link?

On the dashboard and staking page it makes sense to pre-select the network and asset, but it would be nice to let users choose network/asset in the Transak modal when navigating from this button.

src/ui-config/onRampServicesConfig.tsx
```diff
@@ -0,0 +1,23 @@
...
+  name: string;
+  makeLink: ({ cryptoSymbol, network, walletAddress }: MakeLinkParams) => string;
+  icon: ReactNode;
+}
+
+export const onRampServices: OnRampServices[] = [
+  {
+    name: 'Transak',
+    makeLink: ({ cryptoSymbol, network, walletAddress }) =>
+      `${process.env.NEXT_PUBLIC_TRANSAK_APP_URL}/?apiKey=${process.env.NEXT_PUBLIC_TRANSAK_API_KEY}&network=${network}&cryptoCurrencyCode=${cryptoSymbol}&walletAddress=${walletAddress}&disableWalletAddressForm=true`,
```
defispartan, line 20: Could you change `network` to `network.split(' ')[0]`

The Polygon network is called `Polygon POS` in the Aave app which breaks this link, but this change will fix it

src/layouts/MoreMenu.tsx
debajitr: Done

src/ui-config/onRampServicesConfig.tsx
debajitr: Thanks for pointing this out :)
Done

MareskoY:
> @debajitr , Hello, I tested one more time and all is good except header redirection
> 
> it still redirects only to the default Ethereum network and doesn't depend on the market or selected network
> 
> redirection [ redirection ](https://user-images.githubusercontent.com/18032275/194320375-0fe9c72e-34f4-4ac2-a757-3664ea569630.gif) [ ](https://user-images.githubusercontent.com/18032275/194320375-0fe9c72e-34f4-4ac2-a757-3664ea569630.gif)

@defispartan told me that there is some problem over here, and we could deploy the current version

drewcook:
@debajitr Please add in the new environment variables to `.github/actions/build/action.yml` at the end of the file under the `env:` field for the build job.  This is the final step, otherwise this won't work in production.

debajitr:
@drewcook just added, can you please check

.github/workflows/build-test-deploy.yml
```diff
@@ -26,6 +26,9 @@ jobs:
         uses: ./.github/actions/build
         with:
           NEXT_PUBLIC_ENV: 'prod'
+          NEXT_PUBLIC_TRANSAK_APP_URL: 'https://global.transak.com'
+          NEXT_PUBLIC_TRANSAK_API_URL: 'https://api.transak.com'
+          NEXT_PUBLIC_TRANSAK_API_KEY: '799087ea-c2be-44b8-a0fe-56d96b37ee3c'
```
drewcook, line 31: I'm not sure these are totally necessary in this spot.  @defispartan or @RoRu can confirm though

drewcook: This block is saying "use everything in `/build/actions.yml` with these additional variables.  Since they're already added in the original build file, they're not needed here.

drewcook: I'm 99% certain they are redundant here, and can be removed.

RoRu: Can confirm, these won't do anything here, better to remove them

drewcook: Right, it's been a minute, but these can be removed.  If adding in this way, then `actions.yml` would need to take in the dynamic values, and they'd be defined out of scope at this file.  This isn't what we want, so please remove these.

debajitr: Done

drewcook: Great, thanks!

drewcook (approved):
👍

defispartan (approved):
:rocket:

PR #977, opened 2022-07-27
e2e stake-abpt test

bojanaave:
Added a new e2e test for Staking ABP Token

### DIFF
--- a/cypress/integration/2-settings/stake-abpt.ts
+++ b/cypress/integration/2-settings/stake-abpt.ts
@@ -5,14 +5,14 @@ describe('Checking ABPT token modal on Staking page', () => {
     configEnvWithTenderlyMainnetFork({});
 
     it('step1:Open Stake page', () => {
-      cy.get('a[href*="/staking/"]').click();
+      cy.get('[data-cy="menuStake"]').click();
     });
   });
 
   describe('CASE2:GET ABP Token', () => {
     it('step2:Get ABP Token', () => {
       cy.contains('Stake ABPT').click();
-      cy.contains('Get ABP Token').click();
+      cy.get('#getAbp-token').click();
     });
 
     it('step3:Verify does Get ABP Token modal is opened', () => {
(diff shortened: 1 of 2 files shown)


cypress/integration/2-settings/stake-abpt.ts
```diff
@@ -0,0 +1,26 @@
+import { configEnvWithTenderlyMainnetFork } from '../../support/steps/configuration.steps';
+
+describe('Checking ABPT token modal on Staking page', () => {
+  describe('CASE1:Open Stake page', () => {
+    configEnvWithTenderlyMainnetFork({});
+
+    it('step1:Open Stake page', () => {
+      cy.get('a[href*="/staking/"]').click();
```
grothem, line 8: Could target cypress attribute here, that way test wouldn't break if we changed the link name
`cy.get('[data-cy="menuStake"]')`

cypress/integration/2-settings/stake-abpt.ts
```diff
@@ -0,0 +1,26 @@
...
+  });
+
+  describe('CASE2:GET ABP Token', () => {
+    it('step2:Get ABP Token', () => {
+      cy.contains('Stake ABPT').click();
+      cy.contains('Get ABP Token').click();
+    });
+
+    it('step3:Verify does Get ABP Token modal is opened', () => {
+      cy.contains('Go to Balancer Pool').should(
```
grothem, line 19: Same thing, could do
`cy.get('[data-cy="stakeBtn_ABPT"]').click();`

This one would probably be more likely to break. Unlikely that we'd change the link name, but in this case if the button text we're to change, the test would fail :)

cypress/integration/2-settings/stake-abpt.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    it('step1:Open Stake page', () => {
+      cy.get('a[href*="/staking/"]').click();
+    });
+  });
+
+  describe('CASE2:GET ABP Token', () => {
+    it('step2:Get ABP Token', () => {
+      cy.contains('Stake ABPT').click();
+      cy.contains('Get ABP Token').click();
```
grothem, line 15: Looks like we don't have a cypress attribute to target on this button, but we could easily add one. This one would probably be more likely to break. Unlikely that we'd change the link name, but in this case if the button text we're to change, the test would fail :)

bojanaave: Thanks @grothem  , the commit below should fix / implement your suggestions.

grothem: Awesome, looks good

defispartan (approved):

PR #993, opened 2022-08-02
fix: dashboard states

defispartan:
Addresses a few errors:

- Table header showing in 'Assets To Supply' header when the market is frozen and there are no assets

AssetSupplyBug

- Incorrectly displaying "Your wallet balance is empty" on production markets
- On the reserve overview page, the frozen warning for harmony assets is missing a link to the snapshot vote


src/modules/reserve-overview/ReserveConfiguration.tsx
```diff
@@ -156,7 +155,7 @@ export const ReserveConfiguration: React.FC<{ reserve: ComputedReserveData }> =
         <Box>
           <Alert sx={{ mt: '16px', mb: '40px' }} severity="error">
             <Trans>
-              {reserve.symbol} is frozen due to an Aave Protocol Governance decision.{' '}
+              {reserve.symbol} is frozen due to an Aave community decision.{' '}
```
grothem, line 158: I noticed the `FrozenWarning` component has this same text that we might want to make consistent.

On a related note, is there a list of frozen assets somewhere? I was just curious what that warning looked like, but I couldn't find any assets that were frozen.

MareskoY (approved):

src/modules/reserve-overview/ReserveConfiguration.tsx
defispartan: Just created #996 to fix this across multiple places in the app

grothem (approved):

PR #1003, opened 2022-08-04
Fix/950 phantom rewards

grothem:
I believe the root cause for users seeing rewards that don't actually exist is from stale data from the public rpc's configured. This PR will add a higher weight for the private rpc, and also removes the quorum of 1, in favor of the default, which is half the sum of the weights of the providers.

What I was able to reproduce on a fork, was data from a public rpc being used over the private one, because the quorum was set to 1. I also added a condition in the case of forks, to not add any of the configured public rpc's because it doesn't make sense if we're on a fork, it'd be wrong data anyway. With this change, if the responses are different from the private vs. public, the private would take precedence.

It looks like the quorum config was changed recently in this commit: https://github.com/aave/interface/commit/0b6b04f6b379eb1b6c5191204973160a55b48779
so pinging @sakulstra to see if these changes make sense

### DIFF
--- a/src/ui-config/networksConfig.ts
+++ b/src/ui-config/networksConfig.ts
@@ -83,11 +83,7 @@ export const networkConfigs: Record<string, BaseNetworkConfig> = {
   [ChainId.mainnet]: {
     name: 'Ethereum',
     privateJsonRPCUrl: 'https://eth-mainnet.gateway.pokt.network/v1/lb/62b3314e123e6f00397f19ca',
-    publicJsonRPCUrl: [
-      'https://cloudflare-eth.com',
-      'https://rpc.flashbots.net/',
-      // 'https://eth-mainnet.alchemyapi.io/v2/demo',
-    ],
+    publicJsonRPCUrl: ['https://cloudflare-eth.com/v1/mainnet'],
     publicJsonRPCWSUrl: 'wss://eth-mainnet.alchemyapi.io/v2/demo',
     // cachingServerUrl: 'https://cache-api-1.aave.com/graphql',
     // cachingWSServerUrl: 'wss://cache-api-1.aave.com/graphql',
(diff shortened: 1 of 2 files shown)


grothem:
What I still need to test and verify is, that if the private rpc is down, quorum can still be reached with the public rpc

grothem:
Thanks, I updated the network config for forks to handle that case of removing the public rpcs.

I was doing some testing related to the quorum and handling errors, and it seems like if the provider weight is less than the quorum, then it will try the other providers. So I think in our case, we have a quorum set to 1, and if any provider fails then it'll throw an error and not 'fallback' to the other ones. This will also be the case when we have 1 private and 1 public configured for a market, and we use the default quorum calculation.

On my branch, I tried setting a broken url as the private rpc, and left the 2 main net public ones, and it worked as expected.

sakulstra:
@grothem yes the whole situation with fallbackProvider is a bit sad and i'm not even sure if the behavior is a bug or desired - it's more of a quorum provider than a fallback provider. If you dig in git history you'll find a pr where i also tried to apply weight 2 to the private one and then surrendered.

There's multiple issues here:
1. the "FallbackProvider" only falls back on error when the error is an error on initialization (iirc they all call getBlock and ones erroring there will be filtered out)
2. lot's of errors are not reported with an error code, but `200` with an error content (e.g. when an address is not on the allowlist)
3. there are some requests that simply don't work with the cloudflare provider as it does not implement the full spec (iirc in regards to governance)

So putting weight `2` on the private one in some cases leads to both public ones needing to find a quorum which they can't eventually leading to a deadlock. The overall situation is a bit sad.

grothem:
@sakulstra thanks for the additional context. Sounds like we have to make a compromise somewhere because of how the fallback provider currently work unfortunately.

If we go with the default quorum calculation, the only market it would have an impact on is eth mainnet, since all other markets have 1 private and 1 public right now. I think a quorum of 2 would be enough to solve the bug with missing rewards showing up, and now that we're using the private pokt rpc's, hopefully that'll minimize errors we see.

grothem:
So I ended up reverting the change to quorum, and instead I removed one of the public rpc's for mainnet (I also updated it to match cloudflare docs). I think this change would make the phantom rewards less likely to happen, but doesn't completely fix it.

I was talking with @defispartan, and there might be an opportunity to so some refactoring around how we handle the fallback providers when we add in the ability for users to use their connected wallet provider.

MareskoY (approved):

defispartan (approved):
Context: stale data from public rpcs may be causing incorrect values from incentive data contracts. By removing the fallback provider, this may result in less cases

PR #1011, opened 2022-08-08
chore: add in unit testing scaffolding

drewcook:
# General Changes

- adds in `jest` and supporting packages with RTL support
- adds a Nextjs-compatible config
- adds a `test` and `test:ci` command that can be run in CI
- adds an example test to prove the `test` command works effectively
- ignores any coverage testing for now

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,0 +1,17 @@
+/* eslint-disable-next-line @typescript-eslint/no-var-requires */
+const nextJest = require('next/jest');
+
+const createJestConfig = nextJest({
+  // Provide the path to your Next.js app to load next.config.js and .env files in your test environment
+  dir: './',
+});
+
+// Add any custom config to be passed to Jest
+const customJestConfig = {
+  setupFilesAfterEnv: ['<rootDir>/jest.setup.js'],
+  testEnvironment: 'jest-environment-jsdom',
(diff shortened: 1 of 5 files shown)


grothem:
Nice, just one thing to note, once we start adding tests we'll have to get a CI step setup to run those

jest.config.js
```diff
@@ -0,0 +1,17 @@
...
+  // Provide the path to your Next.js app to load next.config.js and .env files in your test environment
+  dir: './',
+});
+
+// Add any custom config to be passed to Jest
+const customJestConfig = {
+  setupFilesAfterEnv: ['<rootDir>/jest.setup.js'],
+  testEnvironment: 'jest-environment-jsdom',
+  testMatch: ['**/__tests__/**/*.[jt]s?(x)', '**/?(*.)+test.[jt]s?(x)'],
+};
```
grothem, line 14: Do we need the `moduleDirectories` config? Looking at their docs, it seems like we do... but seems like it works without it, so maybe not 😅

drewcook: I think you're right.  We do have a baseUrl in `tsconfig.json`.  I'll add it back in.

grothem (approved):

MareskoY (approved):

Issue #732, opened 2022-04-30
Emphasize repay with aToken feature


defispartan:
In V3, borrowed assets can be directly repaid with aTokens. Currently the feature is kind of hidden though. You have to click the arrow next to the asset, then select the aToken. Maybe this could be redesigned to emphasize it a bit more.

![repayWithAToken](https://user-images.githubusercontent.com/26660211/166089563-8f84f1c5-23c2-4413-80a9-e4d0ad57b0c3.png)

defispartan:
A first implementation of this will be unselecting the asset in the "Wallet Balance" column of the Repay modal if the user has more than option, so the user will open the menu and see the option to repay with aTokens in V3.

This will also highlight repayinh with wrapped base tokens (WETH, WMATIC, etc.) in V2 and V3.

helpfulBro:
Had a user today trying to repay with USDC collateral and could not find it in the expected place.

@iamanastasia 
Can I suggest that aToken collateral also be represented in the 'collateral' tab as this is the natural location a user would look for this. 

![image](https://user-images.githubusercontent.com/25915621/201588411-fc8a244b-0753-4652-9dfd-c7402730edfb.png)
![image](https://user-images.githubusercontent.com/25915621/201588430-7145de3c-1f39-48dd-9783-68e6a7552762.png)

PR #1016, opened 2022-08-10
fix: ordering and available amounts in repay modal

defispartan:
- [x] Do not allow negative values in repay modal -  Removed the condition that 0.004 base asset be subtracted. This isn't relevant for most non-ETH networks and will show an error in tx confirmation if this is reached. Additionally, the max balance is confusing for users who think their assets are missing.

### DIFF
--- a/src/components/transactions/Repay/RepayModalContent.tsx
+++ b/src/components/transactions/Repay/RepayModalContent.tsx
@@ -127,8 +127,7 @@ export const RepayModalContent = ({
     // set possible repay tokens
     // if wrapped reserve push both wrapped / native
     if (poolReserve.symbol === networkConfig.wrappedBaseAssetSymbol) {
-      // we substract a bit so user can still pay for the tx
-      const nativeTokenWalletBalance = valueToBigNumber(nativeBalance).minus('0.004');
+      const nativeTokenWalletBalance = valueToBigNumber(nativeBalance);
       const maxNativeToken = BigNumber.max(
         nativeTokenWalletBalance,
         BigNumber.min(nativeTokenWalletBalance, debt)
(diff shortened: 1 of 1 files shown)


src/components/transactions/Repay/RepayModalContent.tsx
```diff
@@ -163,7 +162,7 @@ export const RepayModalContent = ({
         balance: maxBalance.toString(10),
       });
     }
-    setAssets(repayTokens);
+    setAssets(repayTokens.sort((a, b) => Number(a.balance) - Number(b.balance)));
```
grothem, line 165: Would sorting by usd value here instead of balance be better UX?

MareskoY:
there is important bug over here 

Repay action for each base asset doesn't work, repayable amount 0 every time 

Screenshot 2022-08-11 at 12 18 43

Screenshot 2022-08-11 at 12 18 56


p.s. after fix tests should be passed

defispartan:
>

src/components/transactions/Repay/RepayModalContent.tsx
defispartan: The only options in this modal are the underlying asset, the aToken, and a wrapped version of the assets which will all have the same USD price so sorting by balance will have the same affect

defispartan:
> there is important bug over here
> 
> Repay action for each base asset doesn't work, repayable amount 0 every time
> 
> Screenshot 2022-08-11 at 12 18 43
> 
> Screenshot 2022-08-11 at 12 18 56
> 
> p.s. after fix tests should be passed

This is repaying with WETH, if you're trying to repay with ETH you'll have to change the selected asset with the arrow. There may be a different bug that the ordering is reversed (lowest balance to highest balance), I'll check on that

src/components/transactions/Repay/RepayModalContent.tsx
defispartan: Reverted this ordering change. Always showing the base asset first is the best UX. The issue of not being able to find these other options should be addressed separately in #732

defispartan:
@MareskoY I reverted the change to the ordering so it should always default to ETH again and fix the issue in your screenshot.

MareskoY (approved):

grothem (approved):

Issue #836, opened 2022-06-07
Add clear button for amount inputs


defispartan:
![input](https://user-images.githubusercontent.com/26660211/172289142-82dd76c7-9b4a-4bcc-ad84-b1caba0f42a5.PNG)

Currently, it's quite tedious to input a custom amount after selecting the max button. You have to manually move the cursor to the end of the amount and backspace. It would be much nice to have a "clear" or "x" button to reset the amount to 0.

iamanastasia:
Yeah, let's improve to classic interaction with x icon. But also would be great to see abbreviation of long numbers.

<img width="525" alt="Screenshot 2022-07-29 at 17 44 16" src="https://user-images.githubusercontent.com/92739089/181774142-018dcc6e-2c77-4fc1-85b9-ddb9a41b70a9.png">

MatthewCYLau:
👋 Hi there! New contributor here, and this ticket looks like a fantastic newbie ticket.

What is the user journey to render `AssetInput`? I'm keen to pick this one up.

Thanks! 🙏

defispartan:
Hey! Yeah this would be a great first ticket, it's mostly css  contained to one component

`AssetInput` is rendered inside of Modal components where the user is selecting an amount: supply, borrow, repay, withdraw

MatthewCYLau:
👋 Raised a small PR - please let me know what you think 🙏

PR #1027, opened 2022-08-15
Feat: added clear button for asset input

MatthewCYLau:
- Resolves https://github.com/aave/interface/issues/836
- Added clear button for asset input through `CancelIcon`
- Render button when value is present
- On click, clears value

image

### DIFF
--- a/src/components/transactions/AssetInput.tsx
+++ b/src/components/transactions/AssetInput.tsx
@@ -150,7 +150,7 @@ export const AssetInput = <T extends Asset = Asset>({
               {value !== '' && (
                 <Button
                   size="small"
-                  sx={{ minWidth: 0, ml: '7px', p: 0 }}
+                  sx={{ minWidth: 0, ml: '7px', p: 0, left: 8 }}
                   onClick={() => {
                     onChange && onChange('');
                   }}
@@ -162,7 +162,7 @@ export const AssetInput = <T extends Asset = Asset>({
               <TokenIcon
                 aToken={asset.aToken}
                 symbol={asset.iconSymbol || asset.symbol}
-                sx={{ mr: 2, ml: 2 }}
+                sx={{ mr: 2, ml: 4 }}
               />
               <Typography variant="h3" sx={{ lineHeight: '28px' }} data-cy={'inputAsset'}>
                 {symbol}
--- a/src/components/transactions/AssetInput.tsx
+++ b/src/components/transactions/AssetInput.tsx
@@ -2,6 +2,7 @@ import { Trans } from '@lingui/macro';
 import {
   Box,
   Button,
+  IconButton,
   FormControl,
   InputBase,
   ListItemText,
@@ -9,8 +10,9 @@ import {
   Select,
   SelectChangeEvent,
   Typography,
+  useTheme,
 } from '@mui/material';
-import CancelIcon from '@mui/icons-material/Cancel';
+import { XCircleIcon } from '@heroicons/react/solid';
 import React, { ReactNode } from 'react';
 import NumberFormat, { NumberFormatProps } from 'react-number-format';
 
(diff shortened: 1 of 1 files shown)


src/components/transactions/AssetInput.tsx
```diff
@@ -146,6 +147,18 @@ export const AssetInput = <T extends Asset = Asset>({
 
           {!onSelect || assets.length === 1 ? (
             <Box sx={{ display: 'inline-flex', alignItems: 'center' }}>
+              {value !== '' && (
+                <Button
```
grothem, line 151: I think `IconButton` would be better to use here. That'll give us the circle hover/ripple effect, instead of the square.

src/components/transactions/AssetInput.tsx
```diff
@@ -146,6 +147,18 @@ export const AssetInput = <T extends Asset = Asset>({
...
+                  size="small"
+                  sx={{ minWidth: 0, ml: '7px', p: 0, left: 8 }}
+                  onClick={() => {
+                    onChange && onChange('');
+                  }}
+                  disabled={disabled || isMaxSelected}
+                >
+                  <CancelIcon sx={{ fontSize: 16, fill: 'silver' }} />
```
grothem, line 159: We typically use hero icons. I think the `XCircleIcon` from `@heroicons/react/solid` is what you could use.

grothem (commented):
Thanks for the PR! I added a couple of comments.

src/components/transactions/AssetInput.tsx
MatthewCYLau: Fixed - now using `IconButton`

src/components/transactions/AssetInput.tsx
MatthewCYLau: Fixed - now using `XCircleIcon`

grothem:
> * Ipfs hash: bafybeigqhwn2cgzhapwtinp6wx25q6xrrtl6ir5xuato3u2exvseijwp7m
> * Ipfs preview link: https://bafybeigqhwn2cgzhapwtinp6wx25q6xrrtl6ir5xuato3u2exvseijwp7m.ipfs.cf-ipfs.com/

@iamanastasia preview link

MatthewCYLau:
> 

Interesting - it is certainly functionally locally on my development laptop (classic)

My wallet is sadly empty. Is there a way to render Asset Input in the above feature branch environment? I'm keen to poke around myself as well thanks! 🙏

grothem:
@MatthewCYLau you can run the UI against a tenderly fork so you can play around without using real funds. There's a section in the contributing file that explains how to do that: https://github.com/aave/interface/blob/main/CONTRIBUTING.md#run-on-a-fork

You would do those steps after you've created the fork. You should be able to create a free account and make a fork through tenderly https://tenderly.co

Let me know if you run into any issues!

MatthewCYLau:
Thank you, @grothem 🙏 

However, I'm still confused to why this seemingly small button is working fine on my local feature branch, but apparently not clickable on the feature branch environment :sweat_smile:

Would any developer know if I've missed something obvious on how I'm using `IconButton`

grothem:
Oh I see the issue, the button is being disabled if `isMaxSelected`. So if you manually input a number, it works. I think it makes sense to keep the button enabled in that case.

MatthewCYLau:
Okay cool! I've updated condition to disable button. Please let me know if it now works better 🙏

iamanastasia:
@MatthewCYLau @grothem It works! Amazing! Now I don't have to struggle on mobile screens :D 

I wonder can we also add these changes:
1. Change color on hover state. Color style will change from 'Text muted' -> 'Text secondary'.
Screenshot 2022-08-18 at 13 19 49

2. Abbreviate large numbers when input is not active, so it looks more nice.
Screenshot 2022-08-18 at 13 20 07

MatthewCYLau:
@iamanastasia thank you for the feedback 🙏 I've updated the code based on your suggestions. Please let me know what you think 🙏 

- On-hover, secondary text color is used 👇 
image

- Active 👇 
image

- Not active 👇 over-flow text is shown in ellipsis i.e. `...`
image

iamanastasia:
@MatthewCYLau that was fast 🤔 It works, and it looks great, thank you so much!

MatthewCYLau:
Amazing - thank you for the UI review 🎉 

@grothem do let me know if this could be approved, and merged 🙏

grothem (approved):
Thanks! Looks good 👍

defispartan (approved):
Looks good!

PR #1033, opened 2022-08-17
fix: handle screening api 500 errors [skip cypress]

grothem:
Ensure we get a response in the 200 range before trying to parse


foodaka (approved):

defispartan (approved):
Verified this fixes handling of 500 response

PR #1036, opened 2022-08-18
fix: update bridge and faucet button styles

grothem:
Updated style for Bridge and Faucet buttons to match latest design


grothem:
Hold off until #953 is done, need to update styles to match

src/components/FaucetButton.tsx
```diff
@@ -1,24 +1,27 @@
...
   return (
     <DarkTooltip title="Get free assets to test the Aave Protocol">
-      <Button variant="outlined" size="small" onClick={() => router.push(ROUTES.faucet)}>
-        <Typography sx={{ display: 'inline-flex', alignItems: 'center' }} variant="buttonS">
+      <Button component={Link} href={ROUTES.faucet} variant="outlined" size="small">
+        <img
+          src={currentNetworkConfig.networkLogoPath}
+          alt={currentNetworkConfig.name}
+          style={{ width: 14, height: 14 }}
+        />
```
drewcook, line 18: Would something like the `startIcon` prop from MUI's Button work for the image?  Or implementing it similarly to how `TokenIcon` does it, which essentially wraps it in either an `Icon` component.

Simply `<Button startIcon={` maybe or wrapping it with an `Icon` component, or is this over-engineering something as simple as this direct img, which works as is?

src/components/FaucetButton.tsx
```diff
@@ -1,24 +1,27 @@
...
+        <Typography sx={{ mx: 1 }} variant="buttonS">
           <Trans>{currentNetworkConfig.name} Faucet</Trans>
-          <SvgIcon sx={{ fontSize: '14px', mx: '2px' }}>
-            <ExternalLinkIcon />
-          </SvgIcon>
         </Typography>
+        <SvgIcon sx={{ fontSize: 14 }}>
+          <ExternalLinkIcon />
+        </SvgIcon>
```
drewcook, line 24: also with `endIcon` here, respectively

drewcook: I did something similar with `GetABPToken` that worked well with the same look, but I wonder if the use of `component={Link}` would affect these other props

src/components/FaucetButton.tsx
grothem: I tried to use the start and end icon props, but was having trouble sizing the icon correctly. They were too big. Let me take a look at `TokenIcon` and `GetABPToken`, see if there's a cleaner way.

grothem: Updated with to use `startIcon` and `endIcon`. I had to explicitly set the height and width on the icon in order to size it properly.

drewcook (approved):
👍

bojanaave:
LGTM,
Screen Shot 2022-08-24 at 09 58 53
Screen Shot 2022-08-24 at 09 56 58

defispartan (approved):

PR #1051, opened 2022-08-22
change test wallets to hardhat

MareskoY:


cypress/integration/0-v2-markets/4-arc-v2-market/arc-accessibility.arc-v2.spec.ts
```diff
@@ -5,14 +5,16 @@ import {
...
+  privateKey: ``,
 };
 
+//for execute need to add wallet values
 const FULL_WALLET = {
-  address: `0x008c8395eaba2553cde019af1be19a89630e031f`,
-  privateKey: `57bc64f070aeb5ed6f69398ef933e55b74de7a7ec3ebe40f5f009a31a3eff151`,
+  address: ``,
+  privateKey: ``,
 };
```
grothem, line 18: It looks like we could just remove `HALF_WALLET` and `FULL_WALLET`. Where these are being used right now, if we don't provide a value, it'll just use the default wallet.

Maybe we just remove the `wallet` property completely from the `configEnvWithTenderly`, and always use the default test account with a random wallet?

MareskoY: it's not make seanse 

cause arc test don't execute in CI and I keep it here only for future if we will play with arc market again

foodaka (approved):

cypress/integration/0-v2-markets/4-arc-v2-market/arc-accessibility.arc-v2.spec.ts
MareskoY: i just remove arc test =) as suggested by david.k

kartojal (approved):

PR #1053, opened 2022-08-23
feat: enable swap & repayWithCollateral on optimism

sakulstra:


defispartan:
I tested repay with collateral in an optimism fork and it worked fine.

optimism_rwc

My first test of collateral swap resulted in an "out of gas" error on the swap transaction. @bojanaave @MareskoY Could you see if this issue is reproducible? If so, something may need to be added to the gas estimation in aave-utilities.

JurbinM (approved):

defispartan:
Re-tested and collateral swap worked in next two simulated transactions:

optimism_swap

defispartan (approved):

MareskoY:
> I tested repay with collateral in an optimism fork and it worked fine.
> 
> optimism_rwc
> 
> My first test of collateral swap resulted in an "out of gas" error on the swap transaction. @bojanaave @MareskoY Could you see if this issue is reproducible? If so, something may need to be added to the gas estimation in aave-utilities.

I couldn't find this problem I checked the swap around 10 times without any problems, and also verified repay from collateral was also working good 

maybe this was some flaky issue or bug in metamask or somewhere in this area


actually QC passed

MareskoY (approved):

Issue #1050, opened 2022-08-22
Clear Button not visible on Borrow Transaction Modal


0x4Graham:
**Describe the bug**
The "X" icon is not in the borrow transaction modal. This will come from closed change https://github.com/aave/interface/issues/836

![image](https://user-images.githubusercontent.com/7548837/185960939-3cdaed33-e0bf-4f06-b15a-3aa500dae114.png)

**To Reproduce**

1. Go to asset 
2. Select Borrow
3. Enter amount to borrow 
4. No "x" button to clear the amount is visible, 

**Expected behavior**
After entering a value, the "X" button should be visible to clear the input

@MatthewCYLau can you please investigate.

MatthewCYLau:
👋 @grothem @defispartan could I seek you guidance here 🙏 

- Running the latest code from my local machine, I can definitely see the clear button rendering from the `AssetInput` component inside of the `BorrowModalContent` component👇 
![image](https://user-images.githubusercontent.com/25493955/186012203-1188d542-65c4-46f7-b055-6e8abde45f92.png)

- Inside of `AssetInput` component, the clear button renders when `value !== ''` i.e. value is present

- The bug screenshot above is from the `RepayModalContent` component

- And so, I wonder if I've missed anything to how `RepayModalContent` component passes `amount` when rendering `AssetInput` which would result in the clear button not being rendered

Thank you! 🙏

grothem:
Hey @MatthewCYLau, I didn't catch this the first time around, but if the length of the asset list is greater than 1, then the clear button won't show. You can see this if you open the repay modal, and you have more than 1 asset in your wallet that you are able to repay with.

I think another thing that should be added is to not render the `IconButton` if `disabled` is true.

MatthewCYLau:
Raised a small fix PR 🙏

PR #1058, opened 2022-08-24
Fix: fixes clear button not visible on asset input

MatthewCYLau:
- Fixes https://github.com/aave/interface/issues/1050
- Simply render the `IconButton` component i.e. the clear button right after the `InputBase` component
- Update its `z-index` to ensure it is clickable when rendering next to the asset dropdown
- Update its rendering condition to not render if `disabled` is truthy

Screenshot 👇 
image

### DIFF
--- a/src/components/transactions/AssetInput.tsx
+++ b/src/components/transactions/AssetInput.tsx
@@ -148,7 +148,7 @@ export const AssetInput = <T extends Asset = Asset>({
             // eslint-disable-next-line
             inputComponent={NumberFormatCustom as any}
           />
-          {value !== '' && !disabled && (
+          {value !== '' && !disableInput && (
             <IconButton
               sx={{
                 minWidth: 0,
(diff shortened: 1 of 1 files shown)


defispartan (approved):
Looks good!

src/components/transactions/AssetInput.tsx
```diff
@@ -148,28 +148,28 @@ export const AssetInput = <T extends Asset = Asset>({
             // eslint-disable-next-line
             inputComponent={NumberFormatCustom as any}
           />
-
+          {value !== '' && !disabled && (
```
grothem, line 151: This should `!disableInput` instead of `!disable`

`disableInput` is passed in as true in the collateral repay and swap modals.

grothem (commented):
Just one small thing to fix. I should have been more specific on which disabled prop to use 😅

Thanks for taking this on

src/components/transactions/AssetInput.tsx
MatthewCYLau: Fixed 🙏

grothem (approved):

defispartan (approved):

PR #1070, opened 2022-08-30
fix: liquidation risk colors

### LINKED ISSUE #891: Showing incorrect colors in liquidation risk parameters modal

grothem:
If the LTV is grater than the Max LTV, then the line will show as red:
image

I cleaned up some of the logic while in here. I lifted the calculations up to the parent component and then just pass in the alert colors down to the child components.

closes #891


bojanaave:
Looks fine
Screen Shot 2022-08-30 at 17 47 40

defispartan (approved):

drewcook (approved):
👍

PR #1072, opened 2022-08-30
fix: broken svgs and move STABLE config to reserve patches config

sakulstra:


drewcook:
SVGs for buttons and assets are coming through without issues.  The stable coins are on top of list without ordering.  What was broken that's worth checking other than these items?

sakulstra:
@drewcook it's a bit hard to reproduce as it depends on having multiple svgs on the screen with same id and one being hidden.
In most cases the svgs don't break, but just look slightly different.
image

drewcook:
@sakulstra Ah gotcha - okay then, I'll defer to your testing on that. LGTM otherwise! 👍

drewcook (approved):

defispartan (approved):

PR #1073, opened 2022-08-30
feat: add custom fallback provider

sakulstra:
not properly tested at all, also falling back on every error is most definitely wrong


sakulstra:
@drewcook , @defispartan plz have a look and perhaps improve on top of that.

It's a POC of was previously discussed as the current `FallbackProvider` is not what we want.
Instead what we want it a provider that uses a main provider but falls back to another provider on certain errors. I called it `RotationProvider` as it rotates rpcs.

Did some basic tests with breaking providers manually and it seems to kinda work. 
I'd expect though that some finer adjustment on the error handling is necessary. I guess not every error must result in a retry with another provider.

defispartan:
@sakulstra Since this is in the BGD repo I'll create a separate PR for this and pick it up

PR #1082, opened 2022-09-01
fix: use non hardcoded labels as it doesn't scale

sakulstra:
Soon there will be a new eMode added and the current id - name mapping doesn't scale very well as these groups could vary from market to market or network to network.
This pr uses the native eModeLabels

### DIFF
--- a/src/components/transactions/Emode/EmodeModalContent.tsx
+++ b/src/components/transactions/Emode/EmodeModalContent.tsx
@@ -44,18 +44,17 @@ export const EmodeModalContent = ({ mode }: EmodeModalContentProps) => {
   const {
     user,
     reserves,
+    eModes,
     marketReferenceCurrencyDecimals,
     marketReferencePriceInUsd,
     userReserves,
   } = useAppDataContext();
-  const { currentChainId, currentNetworkConfig } = useProtocolDataContext();
+  const { currentChainId } = useProtocolDataContext();
   const { chainId: connectedChainId } = useWeb3Context();
   const currentTimestamp = useCurrentTimestamp(1);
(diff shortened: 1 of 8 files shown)


src/components/transactions/Emode/EmodeModalContent.tsx
```diff
@@ -40,78 +43,43 @@ export interface EmodeModalContentProps {
...
-  const [selectedEmode, setSelectedEmode] = useState<EmodeCategory | undefined>(undefined);
-  const [emodeCategories, setEmodeCategories] = useState<Record<number, EmodeCategory>>({});
-
+  const [selectedEmode, setSelectedEmode] = useState<EmodeCategory | undefined>(
+    getInitialEmode(mode, eModes, user.userEmodeCategoryId)
+  );
   const networkConfig = getNetworkConfig(currentChainId);
 
-  // Create object of available emodes
-  useEffect(() => {
```
sakulstra, line 62: this code was quite literally hell to read and the useEffect was kinda unnecessary

src/components/transactions/Emode/EmodeNaming.ts
```diff
@@ -1,13 +1,6 @@
-export const getEmodeMessage = (categoryId: number, baseAssetSymbol: string): string => {
-  if (categoryId === 0) {
+export const getEmodeMessage = (label: string): string => {
```
sakulstra, line 1: it's probably a bit stupid to keep the function, but was the "least effect" change

drewcook (approved):
These look to be some great optimizations. Everything checks out in the demo URL also 👍

defispartan (approved):
Looks good, much cleaner

PR #1086, opened 2022-09-02
feat: add pinning with 4everland

saullary:
integrate with 4everland hosting service to obtain distributed pin, and global accelerations


saullary:
issue #1085

PR #1089, opened 2022-09-02
feat: eth borrow disable

grothem:
Since ETH borrowing is going to be temporarily disabled, this PR adds in some warnings to let the user know.

The following warnings will be shown when the Ethereum market is selected:

- Warning w/ tooltip for ETH in the borrow positions list
- Warning w/ tooltip for ETH in the market page asset list
- Warning at top of Reserve page for ETH

Once `disableBorrowingOnReserve` is executed from the proposal, the `Borrow` buttons for ETH will get disabled, so there's nothing additional we should need to handle in the UI to disable these buttons.

Proposal: https://snapshot.org/#/aave.eth/proposal/0xa121311c67b7a5bbe5b8b5fe1911663a0ab94ed339a6a4b0e1b9443f670a0e97


src/components/transactions/Warnings/ETHBorrowWarning.tsx
```diff
@@ -0,0 +1,31 @@
+import { ExclamationIcon } from '@heroicons/react/outline';
+import { Trans } from '@lingui/macro';
+import { Box, SvgIcon } from '@mui/material';
+import { ContentWithTooltip } from 'src/components/ContentWithTooltip';
+import { Link } from 'src/components/primitives/Link';
+
+export const ETHBorrowWarning = () => {
+  return (
+    <ContentWithTooltip
```
grothem, line 9: I noticed a bug while testing this with the `ContentWithTooltip` component. Click events aren't prevented from being propagated, so if you click on the link to the in this tooltip, it will also navigate you to the reserve overview page (only on the markets page). I added an issue for this #1093

bojanaave:
Hi @grothem all looks fine except redirection of the "Learn more" link inside tooltip - that "Learn more" link redirects to reserve page - could you fix that? 
_NOTE: Its visible only on mobile screen resolutions_

https://user-images.githubusercontent.com/104987757/188726047-c1a79d75-1a63-40f5-821a-e00f0dc47c05.mov

src/components/transactions/Warnings/ETHBorrowWarning.tsx
grothem: Ended just fixing as part of this PR

grothem:
@bojanaave this is fixed now. This was an issue with all of our tooltips, so might be worth to just double check some others to make sure they still work as before. I went through some and didn't see any issues.

bojanaave:
Hi @grothem there is no warning messages and warning tooltip is not visible anymore.
Screen Shot 2022-09-07 at 10 31 55
Screen Shot 2022-09-07 at 10 31 33

bojanaave (approved):

foodaka (approved):

PR #1094, opened 2022-09-06
feat: add custom error page templates

drewcook:
# Major Changes

- Adds in custom `404.page.tsx`, `500.page.tsx`, and `_error.page.tsx` page components
- If the error is 404, it will show 404 template
- If the error is 500, it will show the 500 template
- If the error is anything other than those two status codes, it will resolve to the 404 template, since that is the logic written in `_error.page.tsx` - however, this template gives the option to expand other templates for other status codes
- Bumps `caniuse-lite` to remove console warning
- Adds `.prettierignore` which should fix some issues

### 404
image

### 500
image


drewcook:
@RoRu These error pages don't resolve in IPFS, but there are some options we have with Cloudflare configs and redirects.  Let's sync on this.

It may or may not be possible to do wildcard redirects in CF to help handle any 404 routes to show our custom error page template, but we'd have to see and dig into CF.

foodaka (dismissed):

foodaka:
@drewcook can you fix the conflicts, everything else looks good to me

drewcook:
@MareskoY Great catch!  It's a great use case for the 404.

I think we should break this feature/bug fix out into a separate task to track it.  It is out of scope for this original PR.  This PR is needed for fixing linting and just adding in the new templates, but nothing functional.  This PR also needs to go out to prevent other branches from having many linting issues and conflicts, so let's break out the bug into a separate issue.

Here's the new issue: https://github.com/aave/interface/issues/1156

foodaka (approved):

PR #1096, opened 2022-09-07
fix: incorrect blocking of rate switch

defispartan:


MareskoY (approved):

defispartan:
Existing logic is correct. The stable rate restriction is as follows:

A user can only switch to a stable rate or borrow at a stable rate if the asset they are borrowing is not being used as collateral and has a higher collateral balance than their borrow amount.

PR #1100, opened 2022-09-07
fix: only show apy info for assets with borrowing enabled

grothem:
For assets that have borrowing disabled, do not show the borrow APY info in the markets assets list

### DIFF
--- a/src/modules/markets/AssetsListItem.tsx
+++ b/src/modules/markets/AssetsListItem.tsx
@@ -72,11 +72,7 @@ export const AssetsListItem = ({ ...reserve }: ComputedReserveData) => {
 
       <ListColumn>
         <IncentivesCard
-          value={
-            reserve.borrowingEnabled || Number(reserve.totalVariableDebtUSD) > 0
-              ? reserve.variableBorrowAPY
-              : '-1'
-          }
+          value={reserve.borrowingEnabled ? reserve.variableBorrowAPY : '-1'}
           incentives={reserve.vIncentivesData || []}
           symbol={reserve.symbol}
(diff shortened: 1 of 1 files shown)


grothem:
@sendra @sakulstra I saw there was just a PR related to this. We were just talking about hiding the APY if borrowing is disabled, regardless of any outstanding debt. Wanted to ping you to see if that makes sense, since I'm removing your recent changes.

sakulstra:
I think it does not make sense :sweat_smile:

When there's still borrowed assets having information about rates is crucial even when borrowing is now disabled.
In the case of wETH this is https://app.aave.com/reserve-overview/?underlyingAsset=0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2&marketName=proto_mainnet crucial as there's 1B$ debt and for the affected ppl current rate is very important and can change quite frequently (when other ppl withdraw).

The change from the morning was done in a rush as ppl needed to see that info.
Would be good to still have an indicator that borrowing is disabled and no new borrows can happen.

src/modules/markets/AssetsListItem.tsx
```diff
@@ -87,7 +83,7 @@ export const AssetsListItem = ({ ...reserve }: ComputedReserveData) => {
       <ListColumn>
         <IncentivesCard
           value={
-            reserve.stableBorrowRateEnabled || Number(reserve.totalStableDebtUSD) > 0
+            reserve.borrowingEnabled && reserve.stableBorrowRateEnabled
```
sakulstra, line 86: good catch (on the miss borrowingEnabled check)

PR #1102, opened 2022-09-08
feat: graph improvements

drewcook:
# General Changes

- Organizing graph components into reusable pieces
  - Adds in `GraphLoading` and `GraphError` components for loading and error states
  - Adds in `GraphTimeRangeSelector` to represent zooming options, set with 1m, 6m, 1y
-  Adds in graph container components, which abstracts them out from the `reserve-overview` page, and allows for displaying and containing data fetching and error/loading states
- Updates the styles of the graph lines, gradients, colors, etc.
- Abstracts out the panel-based components to their own `ReservePanels.tsx` file to clean up the `ReserveConfiguration` component file
- Removes global styling for ToggleButton and made into a re-usable styled component


src/utils/theme.tsx
```diff
@@ -790,48 +794,6 @@ export function getThemedComponents(theme: Theme) {
           },
         },
       },
-      MuiToggleButtonGroup: {
```
grothem, line 793: Some context on this change:

These global styles to the ToggleButton made it difficult to do any one-off customizations of this component, which we needed for the graph time range selector. I removed these styles, and made a re-usable styled component that we can use throughout the app for existing ToggleButtons. Then we could make the necessary modifications in the GraphTimeRangeSelector component.

iamanastasia:
Look great! Especially in dark mode. 

1. Maybe a bug. After I switched dark/light mode couple of times, tooltip got stuck underneath the lines. Not sure why. (probably no one switches modes so often as I do).
Screenshot 2022-09-22 at 15 10 04

2. Need to add % to X-axis for util rate
Screenshot 2022-09-22 at 15 11 49

drewcook:
@iamanastasia I can't reproduce.  But shouldn't affect most users

% has been added

src/components/FadeTransition.tsx
```diff
@@ -0,0 +1,35 @@
...
+    {(state) => (
+      <div
+        style={{
+          ...defaultStyle,
+          ...transitionStyles[state],
+        }}
+      />
+    )}
+  </Transition>
+);
```
drewcook, line 35: So, this isn't actually being used - so open question if we want to keep it in for future use.  I couldn't get it to work all the way with swapping out loading / error for data on the graphs.  In theory, this should fade in/out components based on the parent container's state.

I think a good follow-up to this PR would be to complete this for the graphs, or roll it into any future graph improvements we want to make, like animations.

grothem: My vote would be to just remove it. Seems easy enough to add it back in if/when we add more graph improvements.

drewcook:
Since we don't have the rates history API work ready for v3 markets, we should hide the APY graphs for those markets.  See https://bafybeifbslvfi3zq6qyb7eznpt3y5gwqrj7rq47kog26ujrx6ywzfqusei.ipfs.cf-ipfs.com/reserve-overview/?underlyingAsset=0xd586e7f844cea2f87f50152665bcbc2c279d8d70&marketName=proto_avalanche_v3 as an example.

I suppose our `renderCharts` boolean is not robust enough, as it's literally just checking that the API string value exists.  This needs to have a better check - but I'm not sure what is the best field to inspect or configure for these. 
 There maybe should be another field on the network objects `version: 2|3|5|etc` or similar.
 
 https://github.com/aave/interface/pull/1102/files#diff-ef9202d779738c48b0f0cbce8782a347e82eb2aee4b75f07cc29d90b7e23abd6R39

MareskoY:
@drewcook 

new graphs displaying correct only for Ethereum  v2 and AMM v2 markets, for each another we get error with fetching data 

Screenshot 2022-09-23 at 11 19 30

```
useReservesHistory(): Failed to fetch historical reserve data. 
Object { isCanceled: true }
```

drewcook:
Hiding for v3 just fine now - but the data comes back as an error for v2 polygon and avalanche, due to I think the `from` param on these API calls from `fetchStats` in `useReserveHistory.ts`.  More specifically it could be coming from `resolutionForTimeRange` math.  Albeit, the math works well for Ethereum v2 market.

Here's the same asset from the API (Polygon USDC)

Without the time range:
https://aave-api-v2.aave.com/data/rates-history?reserveId=0x2791bca1f2de4661ed88a30c99a7a9449aa841740xd05e3e715d945b59290df0ae8ef85c1bdb684744

This is the actual call that's happening:
https://aave-api-v2.aave.com/data/rates-history?reserveId=0x2791bca1f2de4661ed88a30c99a7a9449aa841740xd05e3e715d945b59290df0ae8ef85c1bdb684744&from=1661338017&resolutionInHours=6

defispartan:
Disabled charts for Polygon V2 market while we investigate the issue on the API side. This avoids showing a non-functional reload button to users.

@MareskoY Charts should now only be enabled for markets where we have data: ETH V2, ETH AMM V2, Avax V2. Other markets will be enabled as we add API support.

defispartan (approved):
Really great improvements 🤩

bojanaave (approved):
LGTM - checked on desktop / mobile simulators/ and over Iphone 11 Pro Max

drewcook:
Merging this in, and created a new issue for the API issues on Polygon V2, to keep it scoped as a bug.  We know we will make V3 updates in the API.  This feature + bug could be worked on together at the same time, since both will require digging into the API code.

https://github.com/aave/interface/issues/1176

PR #1103, opened 2022-09-08
fix: only show stable borrow rate if borrowing is enabled

grothem:
Adds a check to make sure `borrowingEnabled` is true in addition `stableBorrowRateEnabled` in order to show the borrow apy.

It's possible that stable borrow could still be enabled, but borrowing is disabled on the asset, which takes precedence.


defispartan (dismissed):
@miguelmtzinf confirmed that is a possible scenario in V2, so this is a good fix :+1:

miguelmtzinf (dismissed):

defispartan (approved):

MareskoY (approved):

PR #1105, opened 2022-09-09
fix: switch to using public eth rpc for ens lookups

defispartan:
The primary RPCs used on the app are from POKT and have a customizable contract allowlist. Since it is not feasible to add all ENS resolvers to the allowlist, lookups return an error with the POKT rpc, so this PR switches to using the public rpc for ens lookups.

Addresses to test:

0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045 - vitalik.eth
0x064Bd35c9064fC3e628a3BE3310a1cf65488103D - hello.eth

### DIFF
--- a/src/utils/marketsAndNetworksConfig.ts
+++ b/src/utils/marketsAndNetworksConfig.ts
@@ -150,11 +150,11 @@ const providers: { [network: string]: ethersProviders.Provider } = {};
  * @param chainId
  * @returns provider or fallbackprovider in case multiple rpcs are configured
  */
-export const getProvider = (chainId: ChainId, ens?: boolean): ethersProviders.Provider => {
+export const getProvider = (chainId: ChainId): ethersProviders.Provider => {
   if (!providers[chainId]) {
     const config = getNetworkConfig(chainId);
     const chainProviders: ethersProviders.FallbackProviderConfig[] = [];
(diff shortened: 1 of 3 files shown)


src/utils/marketsAndNetworksConfig.ts
```diff
@@ -150,11 +150,11 @@ const providers: { [network: string]: ethersProviders.Provider } = {};
  * @param chainId
  * @returns provider or fallbackprovider in case multiple rpcs are configured
  */
-export const getProvider = (chainId: ChainId): ethersProviders.Provider => {
+export const getProvider = (chainId: ChainId, ens?: boolean): ethersProviders.Provider => {
```
grothem, line 153: I'm wondering if it'd make sense to just have a new function, `getENSProvider`, instead of adding an optional param to this one. Seems like it's enough of a special case that it might make sense, wouldn't have to have a param for the chainId, just hardcode mainnet and return the public rpc.

defispartan: Makes sense, yeah that would be cleaner

grothem (dismissed):

grothem (approved):

MareskoY (approved):

PR #1106, opened 2022-09-11
feat: add maticx icons

sakulstra:


defispartan:
This still requires an on-chain vote

MareskoY (approved):

defispartan:
Yep, I think that's a good standard to adopt. Lately it has been sporadic where in the governance process the icons are added.

sendra (approved):

PR #1108, opened 2022-09-13
Add label with fork pr pipeline result

marcPineiroAave:
Added a step and action to add a label to all pr from forked repos with a success or failed message to make it easier to filter.


marcPineiroAave:
https://github.com/marcPineiroAave/interface/actions/runs/3058002176 
https://github.com/marcPineiroAave/interface/actions/runs/3058039131
The first one was a test with the pull requests unlabeled. The second one is with a labeled pull request.

The warnings and errors it shows are because the actions is always trying to delete the pass and fail label even if its not there (hence the error), but it's not a stopper, it will leave the pull request with one label with a success or a failure message.

RoRu (approved):
LGTM

drewcook (approved):
👍

PR #1115, opened 2022-09-14
feat: switch governance gateway to cloudflare

defispartan:
Piñata public gateway has been getting frequent errors causing governance cache to not be updated.

This switches to the cloudflare public gateway which should hopefully be more reliable.


foodaka (approved):

defispartan:
> Note CF is also deprecating this sometime this year

Is there a post about this? I did a search and didn't see anything

foodaka:
> > Note CF is also deprecating this sometime this year
> 
> Is there a post about this? I did a search and didn't see anything

@defispartan they notified us on a call

bojanaave:
LGTM

defispartan:
@foodaka This was referring to the specific Cloudflare IPFS hosting service that app.aave.com was previously using, the public gateway should still be maintained

PR #1124, opened 2022-09-14
feat: add interest rate strategy link

defispartan:


src/modules/reserve-overview/ReserveConfiguration.tsx
```diff
@@ -623,6 +625,24 @@ export const ReserveConfiguration: React.FC<ReserveConfigurationProps> = ({ rese
...
+                    </SvgIcon>
+                  }
+                  component={Link}
+                  size="small"
+                  variant="outlined"
+                  sx={{ verticalAlign: 'top' }}
+                >
+                  <Typography variant="buttonS">
+                    <Trans>Interest rate strategy</Trans>
+                  </Typography>
```
drewcook, line 644: Noticed something, is Typography tag necessary here, or should we add it to the other buttons that are similar, like `GetAPBToken`?

Side note on this style button, we've used this a lot in the designs.  At what point do we want to abstract it out into an `ExternalLinkButton`?

defispartan: `buttonS` is actually the default so the `Typography` tag here was unnecessary and I removed it.

And for the second Q I created #1140, that would be a good cleanup.

drewcook (approved):

PR #1133, opened 2022-09-15
chore: ethers upgrade

grothem:
- Upgrade ethers to 5.7.1
- Upgrade @ethersproject/experimental to 5.7.0
- Upgrade @ethersproject/providers to 5.7.1

Changelog:  https://github.com/ethers-io/ethers.js/blob/master/CHANGELOG.md

### DIFF
--- a/cypress/support/tools/bridge.ts
+++ b/cypress/support/tools/bridge.ts
@@ -9,12 +9,15 @@
 
 import { Eip1193Bridge } from '@ethersproject/experimental/lib/eip1193-bridge';
 import { JsonRpcProvider } from '@ethersproject/providers';
-import { providers, Signer, utils } from 'ethers';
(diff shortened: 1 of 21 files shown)


MareskoY (approved):

sakulstra:
> +48.93kb

sounds really bad.

grothem:
> > +48.93kb
> 
> sounds really bad.

Yeah, might be able to cut down on this by being more efficient with our imports. Let me take a look.

sakulstra:
@grothem I think the issue is not shaking or similar, but that the outdated version of @web3-react you're relying on is fixed to 5.5 and gnosis relies on 5.4 so you end up with 3 versions (up from 2, which already wasn't great)

MareskoY:
@grothem pls update with latest `main`

PR #1136, opened 2022-09-15
feat: custom fallback revised

defispartan:
Currently the app does not have a fallback in case of RPC failure, this PR build on #1073 to implement a custom fallback in the form of a `RotationProvider` which works as follows:
- If an rpc returns an error, rotate to next rpc in the array
- If provider rotates away from the first RPC, rotate back after `FALL_FORWARD_DELAY` (1 minute) to prioritize using most reliable RPC
- If provider rotates through all rpcs, wait for `ROTATION_DELAY` (5 seconds) to avoid spamming rpcs with requests


src/utils/marketsAndNetworksConfig.ts
```diff
@@ -153,28 +259,20 @@ const providers: { [network: string]: ethersProviders.Provider } = {};
...
-        })
-      );
+      config.publicJsonRPCUrl.map((rpc) => chainProviders.push(rpc));
     }
     if (!chainProviders.length) {
       throw new Error(`${chainId} has no jsonRPCUrl configured`);
     }
     if (chainProviders.length === 1) {
-      providers[chainId] = chainProviders[0].provider;
+      providers[chainId] = new ethersProviders.StaticJsonRpcProvider(chainProviders[0], chainId);
```
sakulstra, line 273: would recommend to use StaticJsonRpcBatchProvider here as well - forgot to change it

src/utils/marketsAndNetworksConfig.ts
```diff
@@ -143,6 +145,110 @@ export const isFeatureEnabled = {
...
+  private retriedFirst = false;
+
+  constructor(urls: string[], chainId: number) {
+    super(chainId);
+    this.providers = urls.map((url) => new StaticJsonRpcBatchProvider(url, chainId));
+  }
+
+  /**
+   * Add a delay after completing a rotation, so we don't ddos a provider
+   */
```
drewcook, line 190: Can we update/add a comment that mimics what you have in the PR descriptoin for this?  i.e. that if all RPC providers fail we will delay 5000ms and try again, and that's what this is doing.   Also, I think we should store the value as a constant:

`const RETRY_DELAY = 5000`

src/utils/marketsAndNetworksConfig.ts
```diff
@@ -143,6 +145,110 @@ export const isFeatureEnabled = {
...
+    const index = this.currentProviderIndex;
+    try {
+      return await this.providers[index].perform(method, params);
+    } catch (e) {
+      console.log(e.message);
+      await this.rotateUrl();
+      return await this.perform(method, params);
+    }
+  }
+}
```
drewcook, line 250: In general, I think all these class methods warrant a JSDoc comment - will help explain this round robin method a bit for anyone new.

drewcook (changes requested):
Just some commenting - the overall approach looks worthy to try

sakulstra:
> If the first RPC request fails, retry (we want to prioritize the first RPC because this should be the most reliable)

Just out of curiosity, what's the rationale behind this?

> we want to prioritize the first RPC

I feel like you already do, by using it first and as long as it does not break.
Seems a bit weird and like unnecessary complexity to retry directly it if it just failed.

---

When i did the original rpc i was thinking about switching back to the first one after some time automatically (e.g. if there wasn't any error in 1m). Removed it from what i proposed to reduce complexity, but seems more reasonable to me than this approach (as it doesn't hit a currently broken rpc twice in a row).

defispartan:
> > If the first RPC request fails, retry
> 
> Just out of curiosity, what's the rationale behind this?
> 
> > we want to prioritize the first RPC
> 
> I feel like you already do, by using it first and as long as it does not break. Seems a bit weird and like unnecessary complexity to retry directly it if it just failed.
> 
> When i did the original pr i was thinking about switching back to the first one after some time automatically (e.g. if there wasn't any error in 1m). Removed it from what i proposed to reduce complexity, but seems more reasonable & less complex to me than this approach (as it doesn't hit a currently broken rpc twice in a row).

The reason I added this is to add another protection against using the fallback RPCs to prevent splitting requests between RPCs. If a different RPC is used for querying two pieces of data (i.e. incentives and user reserves) you may have conflicting data. If you think this is overkill though I'll remove it

sakulstra:
> If you think this is overkill though I'll remove it

Would remove it, in the end these rpcs should not conflict and if they do it might result in slight ui discrepancies if at all.
Would add the mechanic i described though to "fall forward" back to the first rpc after some time, as i think it's fair to assume that they perform better so ux shouldn't be unnecessarily slowed down.

src/utils/marketsAndNetworksConfig.ts
defispartan: updated

src/utils/marketsAndNetworksConfig.ts
defispartan: Added these as variables in the RotationProvider class

defispartan: Also updated the class description and added comments for individual lines in the rotation logic

src/utils/marketsAndNetworksConfig.ts
defispartan: Added comments to explain how the provider works in detail

src/utils/marketsAndNetworksConfig.ts
```diff
@@ -153,28 +280,20 @@ const providers: { [network: string]: ethersProviders.Provider } = {};
...
-        })
-      );
+      config.publicJsonRPCUrl.map((rpc) => chainProviders.push(rpc));
     }
     if (!chainProviders.length) {
       throw new Error(`${chainId} has no jsonRPCUrl configured`);
     }
     if (chainProviders.length === 1) {
-      providers[chainId] = chainProviders[0].provider;
+      providers[chainId] = new StaticJsonRpcBatchProvider(chainProviders[0], chainId);
```
grothem, line 294: I think the issue related to the cypress test has to do with the new `StaticJsonRpcBatchProvider`. In the cypress setup, we're using the ethers `JsonRpcProvider`.

I'm still digging trying to figure out why it's failing.

grothem: I don't think the tenderly rpcs support batched requests. I swapped out the batch providers for the `StaticJsonRpcProvider` and it worked.

Is the batch provider needed? Or was that just for performance?

src/utils/marketsAndNetworksConfig.ts
```diff
@@ -143,6 +145,143 @@ export const isFeatureEnabled = {
...
+      this.currentProviderIndex += 1;
+    }
+  }
+
+  async detectNetwork(): Promise<Network> {
+    const index = this.currentProviderIndex;
+    try {
+      return await this.providers[index].detectNetwork();
+    } catch (e) {
+      console.log(e.message);
```
drewcook, line 254: `console.error`

drewcook (approved):
One small change for the logging, glad to finally get this one out 👍

src/utils/marketsAndNetworksConfig.ts
```diff
@@ -143,6 +145,134 @@ export const isFeatureEnabled = {
...
+      await this.rotateUrl(index);
+      return this.detectNetwork();
```
grothem, line 247: I'm not sure that this function should be recursive. I think we should only be checking the network for the current provider, and either return or throw. If it throws, I think we'll handle it in the action other action methods and then do the rotate.

grothem: I guess my assumption is is that the base class is calling this when we're trying a provider, and if we switch providers during the middle of that, we'll have multiple retry's going on

src/utils/marketsAndNetworksConfig.ts
```diff
@@ -143,6 +145,134 @@ export const isFeatureEnabled = {
...
+    }
+  }
+
+  // eslint-disable-next-line
+  async send(method: string, params: Array<any>): Promise<any> {
```
grothem, line 252: Is this overriding a base class function? I'm not seeing it in base provider, I think this can go

grothem:
Trying to write some unit tests for this, and just wanted to clarify the behavior:

1. I've got a test that setups 3 providers with bad urls, so they will all fail. It expects that after all 3 are tried, we wait for the `ROTATION_DELAY` time before trying again. Is there any point where we should just throw and stop trying? I think as it is, it'll just keep looping through and try until one succeeds. If we just keep retrying, that may have an impact on our error handling task down the line, maybe just something to keep in mind.
2. My understanding of the `FALL_FORWARD_DELAY` is to prioritize the first one in the list. So if it fails, but the second one succeeds, we'll continue to use the second one (as long as there's no failures) until the `FALL_FORWARD_DELAY` elapses. At that point, it'll rotate back to the first and start the logic over. I think there's a few tests we can write for this piece.

I'm having to make some changes as I get these unit tests setup, so just want to make sure I'm writing the tests correctly.

defispartan:
> Trying to write some unit tests for this, and just wanted to clarify the behavior:
> 
>     1. I've got a test that setups 3 providers with bad urls, so they will all fail. It expects that after all 3 are tried, we wait for the `ROTATION_DELAY` time before trying again. Is there any point where we should just throw and stop trying? I think as it is, it'll just keep looping through and try until one succeeds. If we just keep retrying, that may have an impact on our error handling task down the line, maybe just something to keep in mind.
> 
>     2. My understanding of the `FALL_FORWARD_DELAY` is to prioritize the first one in the list. So if it fails, but the second one succeeds, we'll continue to use the second one (as long as there's no failures) until the `FALL_FORWARD_DELAY` elapses. At that point, it'll rotate back to the first and start the logic over. I think there's a few tests we can write for this piece.
> 
> 
> I'm having to make some changes as I get these unit tests setup, so just want to make sure I'm writing the tests correctly.

1.) As written, no there is not a point where it will stop trying, it will indefinitely continue and wait for the `ROTATION_DELAY` after each full rotation

2.) Yes, that's correct

src/utils/marketsAndNetworksConfig.ts
defispartan: It was intended to help with performance, but during testing it didn't seem to give any improvement. I changed it back to `StaticJsonRpcProvider` in all cases to keep the logic consistent between production and test cases

src/utils/marketsAndNetworksConfig.ts
defispartan: I think we need to dig into where `detectNetwork` is being called to decide what to do here.

My understanding is that the `StaticJsonRpcProvider` does not call this method when performing an action, only on initialization so this could be dangerous if the error is uncaught and a `perform` or `send` is called on the wrong network

src/utils/marketsAndNetworksConfig.ts
defispartan: It's not on the `BaseProvider` class, but it is a function on the `JsonRpcProvider` class

grothem: Ok, so we should be good to get rid of this since we're inheriting from `BaseProvider`

src/utils/marketsAndNetworksConfig.ts
```diff
@@ -143,6 +144,134 @@ export const isFeatureEnabled = {
...
+
+  /**
+   * If provider rotates through all rpcs, delay to avoid spamming rpcs with requests
+   */
+  async delayRotation() {
+    const now = new Date().getTime();
+    const diff = now - this.lastFullRotationTimestamp;
+    if (diff < this.ROTATION_DELAY) {
+      await sleep(this.ROTATION_DELAY - diff);
+    }
```
grothem, line 203: On the first cycle through, we won't actually wait for the `ROTATION_DELAY` because `this.lastFullRotationTimestamp` doesn't get set until _after_ the delay (in `rotateUrl`)

I'm not sure I'm following what this logic is supposed to be doing, would it make more sense if we just did an `await sleep(ROTATION_DELAY)`?

grothem: I updated this to always just wait the rotation delay time and removed the `lastFullRotationTimestamp` tracking

src/utils/marketsAndNetworksConfig.ts
grothem: I was looking at this a little more, it was a different issue that was causing the weird behavior on failing that's fixed now.

`detectNetwork` should only be called on initialization. And since we're using static json rpc providers, I think it'd make sense for this method to just call `detectNetwork` on all of the providers. Maybe we want to copy some of the logic from the `Fallback` provider on how it handles that.

I don't think it makes sense to call `rotateUrl` here. Should just be something like:
```
async detectNetwork() {
  return await this.providers.map(p => p.detectNetwork());
}
```
Might now exactly like that, but we just need to verify that all provider networks have the same configured chainId

sakulstra: sorry somehow missed.
Do however you please, has been a while since i touched this and would need to have a deeper look.
Iirc when you initialize via `StaticJsonProvider(rpc, networkId)` `detectNetwork` should never fail as it just resolves the static networkId inside `this._networkId`.

src/components/__tests__/example.test.ts
```diff
@@ -1,3 +0,0 @@
-it('has a component test', () => {
-  expect(true).toBeTruthy();
-});
```
drewcook, line 3: RIP

src/utils/marketsAndNetworksConfig.ts
```diff
@@ -183,7 +177,7 @@ export const getProvider = (chainId: ChainId): ethersProviders.Provider => {
 export const getENSProvider = () => {
   const chainId = 1;
   const config = getNetworkConfig(chainId);
-  return new ethersProviders.StaticJsonRpcProvider(config.publicJsonRPCUrl[0], chainId);
+  return new StaticJsonRpcProvider(config.publicJsonRPCUrl[0], chainId);
 };
```
drewcook, line 181: Do we want/need a unit test for the `getProvider` function, since it actually is implementing our rotation provider?

src/utils/rotationProvider.ts
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    // Make sure the network matches the previous networks
+    if (
+      !(
+        result.name === network.name &&
+        result.chainId === network.chainId &&
+        (result.ensAddress === network.ensAddress ||
+          (result.ensAddress == null && network.ensAddress == null))
+      )
```
drewcook, line 47: Do we need normalization of the values here or would they add more robustness to the checks?

drewcook (commented):
Tests are passing and I think they're pretty comprehensive 👍 

A couple questions from me, overall non-blocking.

bojanaave:
LGTM

src/components/__tests__/example.test.ts
defispartan: 😢

src/utils/marketsAndNetworksConfig.ts
grothem: Yeah good idea, added 2 new unit tests around this

src/utils/rotationProvider.ts
grothem: Updated, normalized name and ensAddress

grothem (approved):

defispartan:
Looks good!

src/utils/__tests__/marketsAndNetworkConfig.spec.ts
```diff
@@ -0,0 +1,15 @@
...
+
+it('should use a RotationProvider when there are multiple rpc urls configured for a network', () => {
+  const provider = getProvider(ChainId.mainnet);
+  expect(provider).toBeInstanceOf(RotationProvider);
+});
+
+it('should use a StaticJsonRpcProvider when there is only one rpc url configured for a network', () => {
+  const provider = getProvider(ChainId.polygon);
+  expect(provider).toBeInstanceOf(StaticJsonRpcProvider);
+});
```
drewcook, line 15: 🎉

drewcook (approved):

PR #1141, opened 2022-09-16
feat: refactor frozen warnings

defispartan:
- Clean up logic around market and asset based warnings
- Add mappings for Fantom market warnings


src/components/transactions/Supply/SupplyModalContent.tsx
```diff
@@ -217,7 +216,6 @@ export const SupplyModalContent = ({
         poolReserve.symbol === 'AAVE' &&
         isFeatureEnabled.staking(currentMarketData) && <AAVEWarning />}
       {poolReserve.symbol === 'SNX' && !maxAmountToSupply.eq('0') && <SNXWarning />}
-      {currentNetworkConfig.name === 'Harmony' && <HarmonyWarning learnMore={true} />}
```
defispartan, line 220: This warning is unnecessary since supply is disabled when market is frozen

src/components/transactions/Warnings/MarketWarning.tsx
```diff
@@ -0,0 +1,31 @@
...
+  learnMore,
+  warningMessage,
+  linkHref,
+  warningType,
+}: MarketWarningProps) => {
+  return (
+    <Warning severity={warningType}>
+      <Typography variant="caption">
+        <Trans>
+          {warningMessage}{' '}
```
drewcook, line 23: I think these props may need to be passed in with the <Trans> tag at the higher level.  Looking into this a couple months ago exposed that issue, where Crowdin can't work through the strings coming in as props and then being translated, since they could be dynamic.

drewcook: It's not a very scalable system, but with this translation tag, I haven't found another way yet that dynamic translations can happen. I don't think it's even possible though... and the props are handled at runtime.

defispartan: Yep yep, noticed right as I pushed and fixed with latest commit

Also removed some of the unnecessary props here

src/components/infoTooltips/FrozenTooltip.tsx
```diff
@@ -0,0 +1,51 @@
...
+  symbol?: string;
+  currentMarket?: string;
+}
+
+export const getFrozenProposalLink = (
+  symbol: string | undefined,
+  currentMarket: string | undefined
+): string => {
+  if (currentMarket && currentMarket === 'proto_harmony_v3') {
+    return 'https://snapshot.org/#/aave.eth/proposal/0x81a78109941e5e0ac6cb5ebf82597c839c20ad6821a8c3ff063dba39032533d4';
```
drewcook, line 19: Small optimization if you wanted to, assign the baseUrl to a var for DRY

src/components/infoTooltips/FrozenTooltip.tsx
```diff
@@ -0,0 +1,51 @@
...
+  } else if (currentMarket && currentMarket === 'proto_fantom_v3') {
+    return 'https://snapshot.org/#/aave.eth/proposal/0xeefcd76e523391a14cfd0a79b531ea0a3faf0eb4a058e255fac13a2d224cc647';
```
drewcook, line 21: Small optimization if you wanted to, you could assign the baseUrl to a var for DRY

src/components/transactions/Warnings/MarketWarning.tsx
```diff
@@ -0,0 +1,55 @@
...
+    if (forum) {
+      return 'https://governance.aave.com/t/harmony-horizon-bridge-exploit-consequences-to-aave-v3-harmony/8614';
+    } else {
+      return 'https://snapshot.org/#/aave.eth/proposal/0x81a78109941e5e0ac6cb5ebf82597c839c20ad6821a8c3ff063dba39032533d4';
+    }
+  } else if (market === 'Fantom') {
+    if (forum) {
+      return 'https://governance.aave.com/t/arc-aave-v3-fantom-freeze-reserves/9166';
+    } else {
+      return 'https://snapshot.org/#/aave.eth/proposal/0xeefcd76e523391a14cfd0a79b531ea0a3faf0eb4a058e255fac13a2d224cc647';
```
drewcook, line 32: Similarly.  I think we should create a `src/constants/` directory for this.  This is how I've typically seen it in large frontend apps, modularizing the constants.  Was thinking about this as I was writing `CHART_HEIGHT`s this week for the graphs.

drewcook (approved):

src/components/transactions/Warnings/MarketWarning.tsx
defispartan: Also will be covered in other comment ^

grothem: Once we get that file setup, we could setup a mapping of market name to proposal info. Then I think in the asset list components we wouldn't need to do any checking of market name, we just look up in the mapping by market name, if it's empty don't render. Could help clean up some of those if statements in the other pages.

grothem (dismissed):

src/components/transactions/Warnings/MarketWarning.tsx
drewcook: I like this idea

MareskoY (approved):

grothem (approved):

PR #1142, opened 2022-09-16
chore: add prettier commands

drewcook:
## General Changes

- Adds in a `.prettierignore` file
- Adds in new prettier commands for linting and fixing formatting
- Shifts original lint commands to new `lint:code` and `lint:code:fix` commands for the ESLint linting
- Utilized the original commands to do linting for both Prettier + ESLint in one go
- Runs `yarn lint:fix` which fixes some `.github` templates
- Updates our `package.json` `contributors` array and alphabetizes
- Adds in the `license` field to `package.json`


grothem:
Nice! We should update `lint-staged` to use the new commands.

I think we can just replace what's there with `yarn lint:fix`

defispartan (approved):
Looks good, nice additions 👍

grothem (approved):

PR #1152, opened 2022-09-19
fix: btc.b symbol

sakulstra:


src/ui-config/reservePatches.ts
```diff
@@ -44,16 +45,7 @@ export const SYMBOL_MAP: { [key: string]: string } = {
  * With the next version of uipooldataprovider https://github.com/aave/aave-v3-periphery/pull/89 this list can be greatly reduced/removed.
  */
 export const SYMBOL_NAME_MAP: { [key: string]: string } = {
-  AMPL: 'Ampleforth',
```
sakulstra, line 47: These name mapping have been created because there was a bug not returning names on the smart contract.
This bug has been fixed and deployed a while ago, so these mappings are mostly unnecessary now.

I checked the mapping i removed, but there might be more that can be removed.
I wasn't sure with things like teher or stETH as the name we have is wrong′/different and that might be on purpose idk.

¹wrong as in: Tether vs Tether USD, Lido staked ether vs Liquid staked ether,...

MareskoY (approved):

drewcook (approved):

PR #1160, opened 2022-09-21
fix: remove supply APY from interest rate model graph

grothem:
Removes the Supply APY from the interest rate model graph.

Supply APY isn't directly correlated with utilization rate so we can't accurately show it on this graph.


defispartan:
The supply rate can be directly modeled for assets where stable borrow is disabled.

So maybe it makes sense to keep this on the chart, but only for assets with stable borrow disabled.

sakulstra:
It isn't exclusively correlated to utilization, so yes would remove it - probably cleaner than having it and ppl wondering why it doesn't match what they receive.

PR #1215, opened 2022-10-07
fix: made connecting with watch mode on page load more reliable

grothem:
## General Changes

- Fixes an issue where the watch mode only address would not re-populate on page load in some cases

## Developer Notes

The main issue with treating the watch mode address different from other providers, is that the state wasn't getting cleared out. So for example, if you were connected MM, and then you enter in a watch address, we'd still keep around the provider info for MM in local storage. I think that's what was causing some issues on page load, there'd be a race condition with loading the provider and loading the watch mode address.

I did a little refactoring around how the watch mode address is handled. I ended up making a `WatchModeOnlyConnector`. The connector stubs out the required methods and takes care of setting and clearing local storage for the `watchModeOnlyAddress`. Going this route lets us hook into the existing connect and disconnect logic because we're just treating it like another wallet/provider type.

There's still some more refactoring/optimization around the provider connect/disconnect that we can do, but I think this is a good first step.

## Author Checklist

## Reviewer Checklist

- [ ]  New third-party packages, if any, do not introduce potential security threats
- [ ]  There are no CI changes, or they have been OK’d by the devops team
- [x]  QA verification has been completed

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
     "@web3-react/frame-connector": "^6.0.9",
     "@web3-react/injected-connector": "^6.0.7",
     "@web3-react/torus-connector": "^6.2.9",
+    "@web3-react/types": "^6.0.7",
     "@web3-react/walletconnect-connector": "^7.0.2-alpha.0",
     "@web3-react/walletlink-connector": "^6.2.14",
     "d3-array": "^3.2.0",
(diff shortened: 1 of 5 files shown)


defispartan (approved):
Looks good! 

Tested various switches between watch and normal connection, and page refreshes no longer drop connections

defispartan:
@MareskoY Could you take a look at the failing test on Optimism?

MareskoY (approved):

PR #1244, opened 2022-10-13
feat: add in governance voters to proposal UI

drewcook:
## General Changes

- Adds in new react components for voters list within the proposal overview page, including:
  - Container which fetches from the API's `/proposal-top-voters` endpoint
  - A list of combined yaes an nays for top 10 addresses
  - A modal that contains detailed information for lists of _all_ voters, both yaes an nays
  - A list item component that will display Twitter avatars and ENS information if available
- Updates the environment variable for our API base URL from `NEXT_PUBLIC_SCREENING_URL` to `NEXT_PUBLIC_API_BASE_URL` which makes more sense to be general
  - Updates all references
- Bumps some housekeeping packages
  - `husky` and `lint-staged`
  - `jest` and `jest-environment-jsdom`
  - `@visx/mock-data` and `@visx/tooltip`
  - `next@12.1.0 -> 12.1.1`
- Fixes issue with Prettier not being able to parse `.env` files (ignores them)

## Developer Notes

---

## Author Checklist

- [x]  The Github issue has been linked to the PR in the Development section

## Reviewer Checklist

- [x]  End-to-end tests are passing without any errors
- [x]  Code style generally follows existing patterns
- [x]  Code changes do not significantly increase the application bundle size


sakulstra:
super cool, but there's an issue with rounding
showing 1.6M as 2M votes is not good probably makes sense to show exact numbers (and only cut of decimals)

src/modules/governance/proposal/VotersListContainer.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+export const VotersListContainer = (props: VotersListProps): JSX.Element => {
+  const { proposal } = props;
+  const [loading, setLoading] = useState(true);
+  const [error, setError] = useState<null | string>(null);
+  const [voters, setVoters] = useState<VotersData>();
+  const [votersModalOpen, setVotersModalOpen] = useState(false);
+  const isMobile = false;
+
+  const votersUrl = `${process.env.NEXT_PUBLIC_API_BASEURL}/data/proposal-top-voters`;
```
sakulstra, line 52: why add centralization here? You already have all the votes cached in build cache and can fetch the missing ones via events.

src/modules/governance/proposal/VotersListContainer.tsx
```diff
@@ -0,0 +1,138 @@
...
+      } catch (e: unknown) {
+        console.error(e);
+        setError(true);
+      }
+      setLoading(false);
+    };
+    getVoterInfo();
+  }, [proposal]); /* eslint-disable-line react-hooks/exhaustive-deps */
+
+  console.log({ voters });
```
drewcook, line 90: Leaving this here until the db table has fully synced for visibility

src/modules/governance/proposal/VotersListContainer.tsx
defispartan: For this first version, there really isn't a strong case for using the endpoint vs build cache besides caching ENS data, however there are plans to expand this into a full dashboard with user profiles and this endpoint allows for:
- Storing, ENS and Twitter data (via [sybil](https://raw.githubusercontent.com/Uniswap/sybil-list/master/verified.json))
- Storing current voting/proposition power for each user, including delegations
- Storing voting/proposition history for each user

src/modules/governance/proposal/VotersListItem.tsx
```diff
@@ -0,0 +1,81 @@
...
+    : textCenterEllipsis(address, 8, 3);
+
+  // For avatars, the order of precedence is Twitter, then ENS, then default Blockie
+  // If ENS avatar does not exist, then use Blockie
+  const displayAvatar = useBlockie
+    ? makeBlockie(address)
+    : twitterAvatar ?? ensAvatar ?? makeBlockie(address);
+
+  // Don't show any results that come back with zero or negative voting power
+  if (voter.votingPower <= 0) return null;
```
grothem, line 33: I think here we need to look at the `votingPower` on the proposal that they voted on. I think this is the users current voting power.

src/modules/governance/proposal/VotersListContainer.tsx
```diff
@@ -0,0 +1,150 @@
...
+          const [yaes, nays] = await resp.json();
+          // Transform data for UI, sort by highest voting power
+          const yesVoters = yaes.map((v: GovernanceVoter) => ({ ...v, vote: 1 }));
+          const noVoters = nays.map((v: GovernanceVoter) => ({ ...v, vote: 0 }));
+          const votersData = {
+            yaes: yesVoters.sort(sortByVotingPower),
+            nays: noVoters.sort(sortByVotingPower),
+            combined: yesVoters.concat(noVoters).sort(sortByVotingPower),
+          };
+          setVoters(votersData);
```
grothem, line 77: I noticed there were some `any` typings here, would it help to create one type for the server response, and one for the UI data. Then you could type `yeas` and `naes` as the server response type, and then do something like
```
const yesVoters = yaes.map<GovernanceVoter>(v => ({ ...v, vote: 1 }));
const noVoters = nays.map<GovernanceVoter>(v => ({ ...v, vote: 0 }));
```

src/modules/governance/proposal/VotersListItem.tsx
grothem: Oh weird, yeah definitely worth taking another look once we get all synced. I was thinking we'd need to look at the `votingHistory` prop for this proposal to get what power was when they voted.

I _think_ `voter.votingPower` is their current power, so it could be 0 if they don't have any aave

src/modules/governance/proposal/VotersListContainer.tsx
drewcook: There was one case where I put in `unknown[]` for `proposalHistory`.  And line 42 has the definition for `VotersData` where the `yaes` and `nays` are of type `GovernanceVoter[]`, but I failed to add in the types for the transformation statements here.  Great catch!  I'm pushing up the fixes for it.

drewcook: See https://github.com/aave/interface/pull/1244/commits/a88ebf6b858f14b37b47ebb9e8c68b1f81a3ac53

src/modules/governance/proposal/VotersListItem.tsx
drewcook: Ahh that makes sense and could definitely be what was going on here.  Will take a look once resynced and inspect the log statement that contains `voters`.  Could be an easy fix to just `voter.votingHistory.filter(h => h.proposal.id === proposal.id)` at the container level during the transform statements, and so that it may easily be passed into `VoterListItem` and easily consumed.  We'd also alter the type as well.

On that note, yes I think it may make sense to create a type for the server data, and that's exactly what I was doing here.  It needs to be fleshed out fully before merging this PR into `main`, IMO

drewcook: This doesn't quite match up completely.  The voting power on each voting history item is vastly larger than the current voting power.  These are determined in the backend. Do we need to factor in voting weight to reach the correct number?  See recent changes that pull this number and display it: https://github.com/aave/interface/pull/1244/commits/bd5ac66361b1d439d952b8fde91780ef04ffc417

src/modules/governance/proposal/VotersListItem.tsx
```diff
@@ -0,0 +1,83 @@
...
+            alignItems: 'center',
+            maxWidth: 90,
+          }}
+        >
+          <Typography variant="subheader1" color={voter.vote ? 'success.main' : 'error.main'}>
+            {voter.vote ? 'YAE' : 'NAY'}
+          </Typography>
+          <Typography variant="subheader1" color="primary">
+            <FormattedNumber
+              value={voter.proposalVotingPower}
```
defispartan, line 71: `normalize(voter.proposalVotingPower, 18)` imported from math-utils

src/modules/governance/proposal/VotersListContainer.tsx
```diff
@@ -0,0 +1,187 @@
...
+type GovernanceProposalTopVotersResponse = [GovernanceVoter[], GovernanceVoter[]];
+
+export type VotersData = {
+  yaes: GovernanceVoter[];
+  nays: GovernanceVoter[];
+  combined: GovernanceVoter[];
+};
+
+const sortByVotingPower = (a: GovernanceVoter, b: GovernanceVoter) => {
+  return a.votingPower < b.votingPower ? 1 : a.votingPower > b.votingPower ? -1 : 0;
```
defispartan, line 65: Should be `.proposalVotingPower` for each

src/modules/governance/proposal/VotersListContainer.tsx
```diff
@@ -0,0 +1,187 @@
...
+          setError(true);
+        }
+      } catch (e: unknown) {
+        console.error(e);
+        setError(true);
+      }
+      setLoading(false);
+    };
+    getVoterInfo();
+  }, [proposal]); /* eslint-disable-line react-hooks/exhaustive-deps */
```
grothem, line 125: You could add a const for the proposal Id and just add that as a dependency for this `useEffect` to prevent firing off a network request on every re-render.

iamanastasia:
Wow! Looks amazing! Design QA passed :)

src/modules/governance/proposal/VotersListContainer.tsx
drewcook: Correct me if I'm wrong, but if having the page open on an active proposal, and a new vote comes in, then we won't fetch the new voters in a live view.  The user would have to reload the page to see the new votes.  In the next.js page, there is a polling to fetch the live proposal information every 30s.  This is why I went with the overall proposal as the trigger for the `useEffect`, so that voter information could be populated in realtime every 30s.

0x4Graham:
Looks great! Nice job :-)

sakulstra:
I thought i commented somewhere but perhaps i forgot.

The rounding you do on the votes is wrong and misleading.
Showing 2M for 1.519M paints a wrong picture.
Probably should be roundDown instead of halfUp and not abbreviating (don't think 0.x votes matter much, but non decimal shouldn't lose precision imo).

Another thing i noticed and find rather weird is that the vote nowhere links to the onchain vote, but only the user.
I think there's probably a demand for both (e.g. with the wrong 2M it's good to have a chance to check the actual number but also to see how/when this vote was casted).

drewcook:
@sakulstra are you referring to the aggregated value in the vote bar component? The individual user voting power values aren’t being rounded and are precise. Can you post a reference url for the 2M you are referring to?

I agree it would paint the wrong picture and we need to show it accurately.

I did posit the question to the support team and product about where the users should link to given the design. It was discussed to go to their address in etherscan. The design also had external link icon next to the voting power and it was decided to not link it anywhere and remove.

This is still the first design iteration we’re executing on.

sakulstra:
@drewcook they are rounded:
https://bafybeibax7boxuq24oalj262aismguvq3ntvqfmq37yvb46io2si5pwofu.ipfs.cf-ipfs.com/governance/proposal/?proposalId=106
image
top vote is not 2M but ~1.5M

drewcook:
@sakulstra I see, that’s a big then in FormattedNumber component, since all the work here is just plugged into it eventually.

src/modules/governance/proposal/VotersListContainer.tsx
grothem: Yeah, that makes sense

sakulstra:
There's still multiple issues here imo:
- there's still rounding which leads to wrong numbers. Every vote counts so cutting off 10k votes (or even just 4) is not a good idea. Cutting of the decimals is okay imo, but everything else is unnecessary. Especially as there is space.
- mobile view is broken on a lot of pages (i think due to overflow in the underlying content area)

[1 on pixel 5] 
image

MareskoY:
@drewcook the main problem is here https://github.com/aave/interface/pull/1244#issuecomment-1286644280

but there is 1 more problem with UI scroll, close the content 
on desktop and mobile versions:

Screenshot 2022-10-21 at 12 42 13

Screenshot 2022-10-21 at 12 42 43

drewcook:
@sakulstra What mobile device was your screenshot taken from?  My iphone12 mini shows it fine.  The modal is no more than 360 on mobile.

@MareskoY If we go any smaller than we won't be able to fit all the information in such a tiny window.  The native scrollbar is without styling.  To style it even more will take a moderate amount of effort and to tweak it to fit.  The scrollbar will disappear after scrolling.

Again, this is our first iteration on this.  We can mark these items and clean them up on another iteration.  Or, my suggestion is to basicaally abbreviate even more of the names, which almost become obselete at that point, or remove other elements for this particular size.  It is only at around 960px when this will be an issue, then it scales up to 1280 fine.  We're talking about a super small subset of pixels that this would occur visually, and I think it's not perfect, but also not worth the effort to get it perfect at this point, and we can circle back later.  I'm happy to hear others' thoughts though.

cc: @iamanastasia

drewcook:
Ah, I am mistaken that it's only at certain viewports.  I think the path of least resistance is to style the scrollbar to fit outside the container on the righthand side, which would put it neatly in the padding for the container, using negative right margin.  Otherwise, there's just too much information in there to add internal padding to allow it to be _within_ the container, imo.

src/modules/governance/proposal/VotersListContainer.tsx
```diff
@@ -0,0 +1,189 @@
...
+        </Box>
+      </Box>
+    );
+
+  if (!voters || voters.combined.length === 0) return <Box sx={{ mt: 8 }} />;
+
+  return (
+    <Box sx={{ my: 8 }}>
+      {listHeaderComponent}
+      <VotersList compact={mdScreen} voters={voters.combined.slice(0, 10)} sx={{ my: 4 }} />
```
defispartan, line 173: Add `pr: 2` here and it will fix the scrollbar overlap

drewcook: But at 960px it's too much.

drewcook: It's really the only fix we can do without adjusting everything else, so this is the way for now.

drewcook:
@MareskoY Scrollbar issue fixed

sakulstra:
@drewcook did you check the link I sent? There in the description I say pixel 5, but same happens on pixel 6a and quite honestly probably on ever mobile chrome.

The issue is the content of ipfs overflowing which makes the modal on top go crazy. Proper fix is probably not on the modal but on the content overflow below.

drewcook:
@sakulstra yeah, let’s break the words then, but non hyphenated? The issues are those hyperlinks fir sure. But I don’t get the issue with the modal on brave for perspective:

97849159-FDDE-4E1D-8754-637898F71843

drewcook:
Non issue on safari also. iPhone 

FF417728-9AB0-4C03-A58F-A13B8C552EAD

drewcook:
Overflow fix incoming with `word-break: break-word`

bojanaave:
LGTM , I've checked changes in this ticket on : 
real iOS 11 Pro Max 
Pixel 2 , Pixel 5 , Pixel 3 , Pixel 4 (over Chrome simulators)
Example screenshots:
IMG_5364FA4B64F4-1
g)
Screen Shot 2022-10-24 at 12 28 10

grothem (approved):

PR #1302, opened 2022-11-03
fix: maintain query parameters

sakulstra:
- bump visx/tooltip: resolves strict mode error
- maintain current query parameters when adding a new query parameter: issue get's visible when e.g. directly going to https://app.aave.com/reserve-overview/?underlyingAsset=0x7f5c764cbc14f9669b88837ca1490cca17c31607&marketName=proto_optimism_v3
- optimize boot: only set the market from localStorage when it's different to the server rendered one
- "optimize" getAssetCapData: When you declare the function inside, every time the component is rendered the function will be defined again. That said, with this fix applied, the initial page load for this page is still crappy as it returns `null` and does not even render a layout. I don't understand why a context is used here, but also didn't have time to further clean this up.

### DIFF
--- a/src/store/utils/queryParams.ts
+++ b/src/store/utils/queryParams.ts
@@ -2,7 +2,6 @@ import Router from 'next/router';
 
 export const setQueryParameter = (key: string, value: string) => {
   if (typeof window !== 'undefined') {
-    console.log('here');
     Router.push({ query: { ...getAllQueryParameters(), [key]: value } }, undefined, {
       shallow: true,
     });
--- a/src/store/utils/queryParams.ts
+++ b/src/store/utils/queryParams.ts
@@ -2,7 +2,9 @@ import Router from 'next/router';
 
 export const setQueryParameter = (key: string, value: string) => {
   if (typeof window !== 'undefined') {
-    Router.push({ query: { [key]: value } }, undefined, { shallow: true });
+    Router.push({ query: { ...getAllQueryParameters(), [key]: value } }, undefined, {
+      shallow: true,
+    });
   }
 };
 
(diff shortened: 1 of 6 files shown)


src/store/utils/queryParams.ts
```diff
@@ -2,7 +2,9 @@ import Router from 'next/router';
 
 export const setQueryParameter = (key: string, value: string) => {
   if (typeof window !== 'undefined') {
-    Router.push({ query: { [key]: value } }, undefined, { shallow: true });
+    Router.push({ query: { ...getAllQueryParameters(), [key]: value } }, undefined, {
```
sakulstra, line 5: as this runs on app startup the Router.state.query might not be ready yet so using native apis to fetch current parameters

grothem (dismissed):

sakulstra:
@grothem plz don't merge yet, might have found another issue

src/store/utils/queryParams.ts
grothem: You left a `console.log` in

defispartan (approved):

drewcook:
@sakulstra Cool, please add the `Ready for Dev Review` label when it's all cleaned up and ready for a final review :)

src/store/utils/queryParams.ts
```diff
@@ -2,7 +2,10 @@ import Router from 'next/router';
 
 export const setQueryParameter = (key: string, value: string) => {
   if (typeof window !== 'undefined') {
-    Router.push({ query: { [key]: value } }, undefined, { shallow: true });
+    console.log('here');
```
sakulstra, line 5:
```suggestion
```

drewcook (approved):
Good catch, LGTM 👍

grothem (approved):

PR #1324, opened 2022-11-17
Added coverage for the X button and dark mode

bojanaave:
Modified and added new coverage for 
 - Switching language and checking translation
 - Dark mode
 - Opening Markets page and clicking on the "X"


foodaka:
@MareskoY do we need this? can you close out otherwise

PR #1325, opened 2022-11-18
chore: permit modifications

defispartan:
## General Changes

approvepermitflow

- Adds toggle menu for approve and permit selection which is persisted through modals/sessions with localStorage

## Developer Notes

- Moves `tryPermit` flag to decide whether permit is available for a market/asset pair to poolSlice
- The wallet preference for approve vs permit is handled in walletSlice
- Uses https://github.com/aave/aave-utilities/pull/461 to generate default gas estimation for permit actions
- Standardizes permitConfig file and it's usage to always use lowercase

---

## Author Checklist

- [ ]  The base branch is set to `main`
- [ ]  The title is using [Conventional Commit](https://www.conventionalcommits.org/en/v1.0.0/) formatting
- [ ]  The General Changes section has been filled out

- [ ]  The PR is in `Open` state and not in `Draft` mode
- [ ]  The `Ready for Dev Review` label has been added

## Reviewer Checklist


src/components/transactions/Repay/CollateralRepayActions.tsx
```diff
@@ -85,7 +85,9 @@ export const CollateralRepayActions = ({
       sx={sx}
       {...props}
       handleAction={action}
-      handleApproval={() => approval()}
+      handleApproval={() =>
+        approval({ amount: repayAmount, underlyingAsset: poolReserve.aTokenAddress })
```
defispartan, line 89: Need to test this to figure out why the current version is working with blank parameters

sakulstra:
I generally think this is an improvement for wallets not supporting permit.
That said i think the implementation/ux of that is pretty weird.

The path for "doing approval" instead of permit is a `link` ... which acts like a button?
Which at least for me is kinda unexpected and very non-obvious.

Why isn't there just a switch or sth to switch between the two methods?
image

Wdyt?

bojanaave:
LGTM 
Screen Shot 2022-11-23 at 11 45 24

src/components/transactions/Repay/RepayActions.tsx
```diff
@@ -37,12 +37,13 @@ export const RepayActions = ({
   const { currentChainId: chainId, currentMarketData } = useProtocolDataContext();
   const repay = useRootStore((state) => state.repay);
   const repayWithPermit = useRootStore((state) => state.repayWithPermit);
-
+  const tryPermit =
+    currentMarketData.v3 &&
+    permitByChainAndToken[chainId]?.[utils.getAddress(poolAddress).toLowerCase()];
```
grothem, line 42: I noticed this doesn't work on forks because of a different chainId. We could move this to a helper in the store and handle the fork logic in there as well.

src/helpers/useTransactionHandler.tsx
```diff
@@ -134,18 +139,10 @@ export const useTransactionHandler = ({
               txHash: undefined,
               loading: false,
             });
-
-            // set use permit to false to retry with normal approval
-            setUsePermit(false);
```
grothem, line 139: I think there's some cleanup we could do in here now. I don't think we need `usePermit` as state and we can just use the param passed in. It's also being returned from this hook, but doesn't look like it's used anywhere.

src/components/transactions/Repay/RepayActions.tsx
defispartan: Good idea, this belongs in the store anyways

src/components/transactions/Repay/CollateralRepayActions.tsx
defispartan: The reason this works is that these params are only used by permit, added them anyways so it will support in the future

grothem: Would it clean it up if we had a `permit` function and an `approval` function in `useTransactionHandler`? That way it'd be clear when/why these parameters are needed.

sakulstra:
@iamanastasia i see your point, but still kinda disagree :)

The thing is that "approval" vs "permit" is not really a user choice but in the end a wallet choice.

If I'm using a wallet `not` supporting permit, I have to:
- realize the app tries to submit via permit and that there was an error... or nothing happening (@defispartan commented they don't consistently throw errors)
- click a link below the grayed out "supply" button (it's no even visually close to the "non reacting" approve button)

Also the 1 vs 2 clicks argument for me is a bit odd.
- For most users it will still be one click.
- For users using wallets without permit support it should be 1 additional click `once`. If permit doesn't work once, it will never work - the switch value could be stored across sessions. So it's "one click once" vs "never again clicking the big approval button, but always clicking a link below the action"

sakulstra:
@iamanastasia actually one more thing to support my point.

With the current approach (and the suggested one in this pr) - gas estimation will **always** be false for permits as it will always be for approval flow (as it's not clear which path the user is taking before the fact).

defispartan:
> @iamanastasia actually one more thing to support my point.
> 
> With the current approach (and the suggested one in this pr) - gas estimation will **always** be false for permits as it will always be for approval flow (as it's not clear which path the user is taking before the fact).

Gas estimation is only based on the [action](https://github.com/aave/interface/blob/03f839a48245822ed091faa32d1cda2da1cb4c15/src/helpers/useTransactionHandler.tsx#L296), not the approval

sakulstra:
> Gas estimation is only based on the [action](https://github.com/aave/interface/blob/03f839a48245822ed091faa32d1cda2da1cb4c15/src/helpers/useTransactionHandler.tsx#L296), not the approval

I know, that's what i pointed out via dm some time ago. That is why it's always wrong.

On approval flow, you omit the `approval cost` which usually is a substantial part of the txn cost (iirc, aave ui includes it because of that).
On permit flow the gas estimation displayed is for e.g. `supply` not `supplyWithPermit`(which usually is a bit more gas intensive as it includes the permit logic)

defispartan:
> > Gas estimation is only based on the [action](https://github.com/aave/interface/blob/03f839a48245822ed091faa32d1cda2da1cb4c15/src/helpers/useTransactionHandler.tsx#L296), not the approval
> 
> I know, that's what i pointed out via dm some time ago. That is why it's always wrong.
> 
> On approval flow, you omit the `approval cost` which usually is a substantial part of the txn cost (iirc, aave ui includes it because of that). On permit flow the gas estimation displayed is for e.g. `supply` not `supplyWithPermit`(which usually is a bit more gas intensive as it includes the permit logic)

Added the approval tx to gas estimation in the last commit. The issue with doing estimations for `supplyForPermit` is that the estimation will fail unless you provide a valid signature.

It would be possible to hide the estimation until the user signs or switch the estimation from `supply` -> `supplyWithPermit` after they sign, but this retriggers the tx building useEffect and causes loading/stuttering in the modal which imo is worse UX than just substituting in the supply gas cost (omitting the approval) since these estimations were within the margin of error of gas price movements from my testing.

iamanastasia:
> 
@sakulstra Do you think this switcher might be better? 
Screenshot 2022-11-29 at 13 16 24

Can it be a config button with a dropdown?
Screenshot 2022-11-29 at 13 29 38

sakulstra:
@defispartan 
> Added the approval tx to gas estimation in the last commit. The issue with doing estimations for supplyForPermit is that the estimation will fail unless you provide a valid signature.

It's exactly the same for supply, gas estimation will fail when there's no approval. Check aave-utilities, there's default gasLimits for e.g. `supply` to have a estimation in these cases. Probably the one for `supplyWithPermit` is missing.


@iamanastasia 
I think both designs are good (personally i'd prefer the toggle, as it's just two options, so hiding behind a settings wheel feels a bit overkill).
Not sure sure about the text, but also don't have a better suggestion. Perhaps the button could say "Sign DAI approval" when permit is enabled and "Approve DAI" if not ? - not sure if that makes it clearer.

grothem:
I'm wondering if the toggle switch will be confusing. Would the user expect some immediate action on toggle? It might not be very clear at first that the toggle will change the function of the 'approve' button.

iamanastasia:
@sakulstra I like the idea that we can store user's choice, but still think that need to play around with designs and wording.

iamanastasia:
This looks to me as a better version than hyperlink at the bottom of the screen. 
Screenshot 2022-11-30 at 17 39 23


I would recommend to either:
1. Publish current version and create new task to improve. 
2. OR Make changes to UI as shown on the screenshot above, and return to that issue later. 


@sakulstra @defispartan @grothem Thoughts? suggestions?

grothem:
I think this looks really good! One question on the tooltip though, will it be confusing to have that inside the button? Would having it next to the fly out menu link be a better spot?

defispartan:
I like it as well, and this new approach will allow us to always generate accurate gas estimations for whichever path the user selects so I think it's worthwhile to do this refactor before shipping

iamanastasia:
@grothem yeah, good point. I used uniswap as reference for that. And since we don't have much space for 'Why do I need to approve(?)', I would put it in the button, it's hard to find in dropdown menu.

iamanastasia:
> @iamanastasia not so sure about the wording (don't have a better suggestion though). Besides that i think it's good.
> 
> The looks a bit weird to me that tx gas cost is no longer attached to the txn buttons. Wouldn't it both fit into the same line?, might be just my personal taste.

There is another branch (repay with collateral). And there is Slippage tolerance near gas price. 
Screenshot 2022-12-08 at 18 35 31

iamanastasia:
great job! @defispartan

bojanaave:
LGTM
Screen Shot 2022-12-14 at 12 12 35
Screen Shot 2022-12-14 at 12 11 32
Screen Shot 2022-12-14 at 12 11 16

src/components/infoTooltips/ApprovalTooltip.tsx
```diff
@@ -0,0 +1,19 @@
+import { Trans } from '@lingui/macro';
+
+import { Link } from '../primitives/Link';
+import { TextWithTooltip, TextWithTooltipProps } from '../TextWithTooltip';
+
+export const ApprovalTooltip = ({ ...rest }: TextWithTooltipProps) => {
+  return (
+    <TextWithTooltip {...rest}>
+      <Trans>
+        To continue you need to grant Aave smart contracts permission to move your funds from your
```
drewcook, line 10: `To continue, you` (add comma)

src/components/transactions/Repay/RepayActions.tsx
```diff
@@ -34,15 +31,17 @@ export const RepayActions = ({
   blocked,
   ...props
 }: RepayActionProps) => {
-  const { currentChainId: chainId, currentMarketData } = useProtocolDataContext();
-  const repay = useRootStore((state) => state.repay);
-  const repayWithPermit = useRootStore((state) => state.repayWithPermit);
+  const [repay, repayWithPermit, tryPermit] = useRootStore((state) => [
```
drewcook, line 34: This could be simplified to `const { repay, repayWithPermit, tryPermit } = useRootStore()`?

src/components/transactions/FlowCommons/RightHelperText.tsx
```diff
@@ -14,32 +18,50 @@ const ExtLinkIcon = () => (
   </SvgIcon>
 );
 
-export const RightHelperText = ({ approvalHash }: RightHelperTextProps) => {
+export const RightHelperText = ({ approvalHash, tryPermit }: RightHelperTextProps) => {
+  const [walletApprovalMethodPreference, setWalletApprovalMethodPreference] = useRootStore(
+    (state) => [state.walletApprovalMethodPreference, state.setWalletApprovalMethodPreference]
+  );
```
drewcook, line 24: Similarly, could be simplified

src/components/transactions/FlowCommons/ApprovalMethodToggleButton.tsx
```diff
@@ -0,0 +1,93 @@
...
+          }}
+        >
+          <ListItemText primaryTypographyProps={{ variant: 'subheader1' }}>
+            <Trans>{ApprovalMethod.APPROVE}</Trans>
+          </ListItemText>
+          <ListItemIcon>
+            <SvgIcon>{currentMethod === ApprovalMethod.APPROVE && <CheckIcon />}</SvgIcon>
+          </ListItemIcon>
+        </MenuItem>
+      </Menu>
```
drewcook, line 90: Could these menu item components be simplified? They look identical other than the `ApprovalMethod.VALUE`

drewcook (commented):
In all the `useRootStore(state => ...)`, much like we do with our other store slices, using an object destructrue syntax can simplify it and just could use `useRootStore()`.

src/components/transactions/FlowCommons/ApprovalMethodToggleButton.tsx
defispartan: Would end up with more code to split to separate component and pass currentMethod, buttonMethod, setMethod, and handleClose as props so I think it's fine to leave as is

src/components/transactions/FlowCommons/RightHelperText.tsx
defispartan: Done

src/components/transactions/Repay/RepayActions.tsx
defispartan: Done

src/components/infoTooltips/ApprovalTooltip.tsx
defispartan: Done

src/helpers/useTransactionHandler.tsx
```diff
@@ -37,19 +49,21 @@ export const useTransactionHandler = ({
...
   const [signatureDeadline, setSignatureDeadline] = useState<string>();
   const signERC20Approval = useRootStore((state) => state.signERC20Approval);
 
   const [approvalTx, setApprovalTx] = useState<EthereumTransactionTypeExtended | undefined>();
   const [actionTx, setActionTx] = useState<EthereumTransactionTypeExtended | undefined>();
+  const [usePermit, setUsePermit] = useState(false);
   const mounted = useRef(false);
+  const walletApprovalMethodPreference = useRootStore(
+    (state) => state.walletApprovalMethodPreference
+  );
```
drewcook, line 66: `const { walletApprovalMethodPreference } = useRootStore()`

defispartan: Done

grothem (approved):

drewcook (approved):

PR #1458, opened 2023-01-31
fix: increase credit delegation permit

satanworker:
Increase credit delegation permits from 30 min of borrow APY to 1 week


src/store/v3MigrationSelectors.ts
```diff
@@ -518,13 +518,13 @@ const addPercent = (amount: string) => {
   return convertedAmount.plus(convertedAmount.div(1000)).toString();
 };
 
-// adding  30 min of variable or either stable or variable debt APY similar to swap
+// adding  7 days of either stable or variable debt APY similar to swap
```
grothem, line 521: Is this comment still accurate? Not sure if it's the same as what we do with the swap logic. I think the spot your referencing is this https://github.com/aave/interface/blob/main/src/hooks/paraswap/common.ts#L230

The other file referenced doesn't exist anymore.

MareskoY (approved):

src/store/v3MigrationSelectors.ts
satanworker: sorry for late reply, I've updated the comment

PR #1534, opened 2023-03-23
fix: collateral swap hf and swap target collateralization details

### LINKED ISSUE #1533: URGENT agEUR health factor not updated yet
USDT needs to be isolated and when it's not it cannot be used as collateral. Thus, the health factor looks correct:
Screen Shot 2023-03-22 at 3 26 59 PM
That being said, I'd do a check to see if USDT would be isolated on the swap or not and change
`can be collateral        yes -> yes` to `can be collateral        yes -> no` accordingly.
Inversely, agEUR cannot be used as collateral right now because of obvious reasons- which it shows on the UI. However, the health factor is still what it would be if it ***was*** being used as collateral
Screen Shot 2023-03-22 at 3 25 22 PM
so `health factor       1.03 -> 1.04` should be `health factor       1.03 -> 0.97`.

unless I am totally wrong and there's something I'm not seeing- fix!

grothem:
## General Changes

- Fixes a bug in displaying the future health factor when the value would be 0. 
- Improves the `Can be collateral` section of the swap modal details section. Previously, we were only checking if the reserve was enabled to be used as collateral, but this doesn't take into account whether a user is in isolation mode or if the user has disabled an asset to be used as collateral. Now, the states should be updated to reflect the current status of the swap source and swap target reserves.
- Made the collateral states in the supply modal and the swap modal consistent
- Fixes #1533 

---

## Author Checklist

## Reviewer Checklist


defispartan (dismissed):

grothem:
Link T-8802

MareskoY:
found just one point which good to fix 

in case if we swap from unavailable to any 

"unavailable" is not red, not sure that it important, 

Screenshot 2023-03-28 at 09 05 18

grothem:
> found just one point which good to fix
> 
> in case if we swap from unavailable to any
> 
> "unavailable" is not red, not sure that it important,
> 
> Screenshot 2023-03-28 at 09 05 18

Currently it's only shown in red text when the reserve has been disabled from being used as collateral at the protocol level. Which is different from being disabled by the user, or if the user is in isolation mode (in that case it'll be in grey text). This matches how the supply modal works, but maybe it's confusing and we should just use one color? @iamanastasia what are your thoughts on that

iamanastasia:
@grothem good point. Let's make 'unavailable' red as well. 

Question, in supply modal we have an exclamation icon. Let's remove it. Red color is enough to highlight danger zone. 
Screenshot 2023-03-28 at 18 44 23

defispartan (dismissed):

grothem:
Thanks @iamanastasia, good catch. I'll update it

defispartan (approved):

src/components/isolationMode/IsolatedBadge.tsx
```diff
@@ -1,39 +1,115 @@
...
+};
+
+export const IsolatedDisabledBadge = () => {
+  return (
+    <ContentWithTooltip
+      tooltipContent={
+        <IsolationModeTooltipTemplate
+          content={
+            <Trans>
+              Asset can be only used as collateral in isolation mode with limited borrowing power.
```
foodaka, line 65: a bit wordy - wdyt @grothem @0x4Graham

src/components/transactions/utils.ts
```diff
@@ -29,3 +33,56 @@ export const zeroLTVBlockingWithdraw = (user: ExtendedFormattedUser): string[] =
   });
   return zeroLTVBlockingWithdraw;
 };
+
+export const getAssetCollateralType = (
```
foodaka, line 37: Definitly better having this here

foodaka (approved):

Issue #1156, opened 2022-09-19
Redirect to 404 page if reserve details asset is invalid


drewcook:
**Describe the bug**
A clear and concise description of what the bug is, including the exact market(e.g. polygon v3) on which it was observed.

in case we use the link with invalid params:
like incorrect market value or asset address we get a white screen

example:

https://app.aave.com/reserve-overview/?underlyingAsset=0xd586e7f844cea2f87f50152665bcbc2c279d8d70&marketName=proto_avala

We have to use 404 page in this case.

drewcook:
Typically, I'd do this via `getServerSideProps`, inspect the query param, make the request to get the details, and display prior to page load.  But because we build a static site, this may behave differently, and we'll probably want to use `getStaticProps`.  We should do a redirect to the `/404/` page if it's not a valid value.

Jdecristi:
@drewcook I have a solution for this fix, and would like to take this issue.

Solution Proposal:

Implementing getStaticProps will run at build time rather than query. To handle redirection based on route parameters, redirect on static. This can be achieved by checking for the existence of a reserve based on the underlyingAsset during page mount. If the reserve exists, load the page. Otherwise, redirect to the 404 page.

foodaka:
Closing out due to inactivity

PR #1551, opened 2023-04-13
Fix: Redirect to 404 page if reserve details asset is invalid

Jdecristi:
## General Changes

- Fixes [#1156](https://github.com/aave/interface/issues/1156)

## Developer Notes

Implementing getStaticProps will run at build time rather than query. To handle redirection based on route parameters, redirect on static. This can be achieved by checking for the existence of a reserve based on the underlyingAsset during page mount. If the reserve exists, load the page. Otherwise, redirect to the 404 page.

---

## Author Checklist

- [x]  The Github issue has been linked to the PR in the Development section

## Reviewer Checklist

### DIFF
--- a/pages/reserve-overview.page.tsx
+++ b/pages/reserve-overview.page.tsx
@@ -19,24 +19,30 @@ import { ContentContainer } from '../src/components/ContentContainer';
 export default function ReserveOverview() {
   const router = useRouter();
   const { reserves } = useAppDataContext();
-  const underlyingAsset = router.query.underlyingAsset as string;
   const { breakpoints } = useTheme();
   const lg = useMediaQuery(breakpoints.up('lg'));
 
+  const [reserve, setReserve] = useState<ComputedReserveData>();
   const [mode, setMode] = useState<'overview' | 'actions' | ''>('');
+
+  const underlyingAsset = router.query.underlyingAsset as string;
 
   useEffect(() => {
     if (!mode) setMode('overview');
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [lg]);
 
-  const reserve = reserves.find(
-    (reserve) => reserve.underlyingAsset === underlyingAsset
-  ) as ComputedReserveData;
+    if (router.isReady && reserves.length) {
+      const reserveFound = reserves.find((reserve) => reserve.underlyingAsset === underlyingAsset);
+
+      if (reserveFound) setReserve(reserveFound);
+      else router.replace('/404');
+    }
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [lg, router.isReady, reserves.length]);
 
   const isOverview = mode === 'overview';
 
-  return (
+  return reserve ? (
     <AssetCapsProvider asset={reserve}>
       <ReserveTopDetails underlyingAsset={underlyingAsset} />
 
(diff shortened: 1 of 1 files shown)


pages/reserve-overview.page.tsx
```diff
@@ -22,21 +22,28 @@ export default function ReserveOverview() {
...
+      const reserveFound = reserves.find(
+        (reserve) => reserve.underlyingAsset === underlyingAsset
+      ) as ComputedReserveData;
+
+      if (reserveFound) setReserve(reserveFound);
+      else router.replace('/404');
+    }
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [lg, router.isReady, reserves]);
```
defispartan, line 42: @grothem @JoaquinBattilana Do you think it makes sense to change the `reserves` dependency here to `reserves.length` so this useEffect doesn't re-render on every `useCurrentTimestamp` update?

grothem: Sorry, just getting around to take a look at this. For the dependency, could we just use the underlying asset?

I think the overall approach makes sense, just a couple of suggestions:

```ts
if (router.isReady && reserves.length) {
...
}
```
I think this could be simplified, and only check if the reserves are loading. The `useAppDataContext` hook has a `loading` property that you could use here. I think the router will always be ready before the reserve data is loaded so it's redundant.

```ts
const reserveFound = reserves.find(
   (reserve) => reserve.underlyingAsset === underlyingAsset
) as ComputedReserveData;
```
I know this was existing, but it'd be awesome to get this cleaned up. Now that we know the reserves are loaded, we can get rid of the `as ComputedReserveData` type assertion.

The other thing I noticed that this change causes is that we'll lose some of our loading states in some of the child components because we were always rendering before, it's just that sometimes the reserve was undefined. I think we probably need to refactor some of that loading logic, but that'd probably be a different task.

Thanks for taking a look at this @Jdecristi !

Jdecristi:
@defispartan @grothem I fixed the issues you mentioned

```
  const { reserves, loading } = useAppDataContext();

  ...

  useEffect(() => {
    if (!mode) setMode('overview');

    if (loading && reserves.length) {
      const reserveFound = reserves.find((reserve) => reserve.underlyingAsset === underlyingAsset);

      if (reserveFound) setReserve(reserveFound);
      else router.replace('/404');
    }

    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [lg, loading, reserves.length]);
```

defispartan:
The re-direct to 404 is working, but when I click an asset from the markets or dashboard page it just takes me to a blank reserve page and I have to refresh for the page to load. This happens in development and production mode

Jdecristi:
@defispartan I figured out the issue,  it was when I changed the if ``` (router.isReady && reserves.length) ```  to ```if (loading && reserves.length)``` the loading does check the reserves.  However,  it doesn't check on re-render. I had to change it back to fix the issue.

foodaka:
@Jdecristi if I visit the reserve page such as DAI. And then I change the query params to something like https://app.aave.com/reserve-overview/?underlyingAsset=213&marketName=proto_mainnet_v3 I still do not get a 404 page. Can you try and fix? Thank you for your contributions!!

JoaquinBattilana:
Old, not relevant atm.

PR #1733, opened 2023-08-10
Fix/ipfs ci change

alexmance:
## General Changes

Separate workflow to allow different behavior depending of env.

Currently the file build-test-deploy worked indistinctly for PRs
or main. Separating them will allow us better handling of which 
IPFS is used to store development or production releases.


.github/workflows/build-test-deploy-dev.yml
```diff
@@ -0,0 +1,272 @@
...
+          NEXT_PUBLIC_MIXPANEL: ${{ secrets.NEXT_PUBLIC_MIXPANEL }}
+          NEXT_PUBLIC_FIAT_ON_RAMP: 'false'
+          NEXT_PUBLIC_WALLET_CONNECT_PROJECT_ID: ${{ secrets.WALLET_CONNECT_PROJECT_ID }}
+
+      - name: Upload artifacts
+        uses: ./.github/actions/upload-artifacts
+        with:
+          BUILD_ARTIFACT_NAME: 'out'
+
+  build_staging:
```
grothem, line 39: I think for this workflow we wouldn't need the `build_staging` job.

foodaka: yeah agreed, we don't use staging really so we can remove this

PR #1904, opened 2024-01-11
feat: add network to voiting info

MareskoY:
## General Changes

Screenshot 2024-01-11 at 15 33 23


src/modules/governance/proposal/VoteInfo.tsx
```diff
@@ -42,9 +47,45 @@ export function VoteInfo({ proposal }: VoteInfoProps) {
...
+            {network && (
+              <Box
+                sx={{
+                  display: 'flex',
+                  alignItems: 'center',
+                  color: 'text.secondary',
+                }}
+              >
+                <Typography variant="caption">
+                  <Trans>This vote is free on</Trans>
```
grothem, line 66: We currently are not using the relayer for gasless voting. Maybe for now we could just change this text to `Voting is on` and then we can update later.

grothem (changes requested):

src/modules/governance/proposal/VoteInfo.tsx
foodaka: @MareskoY can you update we can get this into main

foodaka:
@MareskoY can u fix the conflicts

foodaka (changes requested):
needs to fix conflicts

foodaka (approved):

grothem (approved):

PR #2029, opened 2024-04-11
fixes: gho bridge

grothem:
- Made a `useGetBridgeMessage` hook that'll return the message and the fee. This logic was just moved from the modal into this hook.
- I noticed that the modal content was rendering immediately on app load, so I refactored a little to match the existing modals. Now, it won't render until the modal is opened. So now there's a `BridgeModal` and a `BridgeModalContent`. This is why the diff is so ugly 😔
- Fixes a bug where the wrong network was being used to estimate gas and so it would fail on approval transactions in some cases
- Some UX improvements with showing loading states when appropriate.
- Fixes USD amount not showing


src/components/transactions/Bridge/useGetBridgeMessage.tsx
```diff
@@ -0,0 +1,143 @@
...
+import debounce from 'lodash/debounce';
+import { useEffect, useMemo, useState } from 'react';
+import { useRootStore } from 'src/store/root';
+import { getProvider } from 'src/utils/marketsAndNetworksConfig';
+
+import { MessageDetails, TokenAmount } from './BridgeActions';
+import { getRouterConfig } from './Router';
+import routerAbi from './Router-abi.json';
+
+export const useGetBridgeMessage = ({
```
foodaka, line 12: this is a really nice optimization

foodaka (approved):

PR #2061, opened 2024-05-06
feat: gho bridge tx history

grothem:


src/components/transactions/Bridge/BridgeActions.tsx
```diff
@@ -185,44 +186,19 @@ export const BridgeActions = React.memo(
           }
         );
 
-        const onRampInterface = new utils.Interface(onRampAbi);
+        // const onRampInterface = new utils.Interface(onRampAbi);
 
         const receipt = await sendTx.wait(1);
+        console.log('receipt', receipt);
```
foodaka, line 192: should we remove this and these comments if we are not using

src/components/transactions/Bridge/BridgeConfig.ts
```diff
@@ -0,0 +1,124 @@
...
+};
+
+export enum MessageExecutionState {
+  UNTOUCHED = 0,
+  IN_PROGRESS,
+  SUCCESS,
+  FAILURE,
+}
+
+// TODO: make testnet and mainnet config
```
foodaka, line 22: yeah can't forget also for supported networks as well

src/hooks/useBridgeTransactionStatus.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+      return MessageExecutionState.UNTOUCHED;
+    },
+    queryKey: ['executionState', chainId, sequenceNumber],
+    enabled: offRamps?.length > 0,
+    refetchInterval: (data) => {
+      if (data === MessageExecutionState.SUCCESS) {
+        return false;
+      }
+
```
foodaka, line 48:
```suggestion
// Note: 60 seconds
```

src/components/transactions/Bridge/BridgeConfig.ts
grothem: Yeah, I added a function in this file, `getSupportedSourceChains`, it just returns all of the chain ids that are in the config. So essentially if it's not in the lane config, it's not a supported network.

MareskoY (approved):

PR #2092, opened 2024-06-21
feat: implement underlying token APYs

MartinGbz:
## General Changes

- Adds underlying tokens APY to:
  - Dashboard page in "Supply" & "Borrow" sections
  - Markets page
- Desktop & Mobile are managed
- Only the sum of both APYs is displayed, a tooltip button has been added to see the APY details (Supply/Borrow APY & Underlying APY)
- APY are fetched/compute using onchain data. But if the onchain fetch failed (eg: no events during the last week), there is for each token a backup function that use an API (except for the `cbETH` that returns 0 because I didn't found any official API, so no underlying APY will be display in this case).

  Q: what are the chances (no events during the last week) ?
  A:
  - sDAI => null, cause we do a contract call
  - stETH, ETHx, weETH, cbETH, rETH => almoast 0, based on the event history, there is always 1 "rebase" event per day (except for weeETH that has 4 event per day), and it's been going on for a long time. If you want I can provide you the list of Rebase event of eacj LST so you can check by yourself.

  But event if one day is 1 day without event, I take the last and latest event from a period of 1 week, so the front will use backup function only if there are no events during the last 7 days. But if you want we can increase it for 14 days to be full safe, I'm just afraid about RPC range block error / rate limit.

### APYs Implemented:
- [x] wstETH
- [x] weETH
- [x] sDAI
- [x] rETH
- [x] ETHx
- [x] cbETH

## Developer Notes

I didn't implement the APY in the migration page because I think it's clearer to keep only Aave Supply APYs here.

For the fork mode, I set the max block range (for event logs fetching) to 1k. Because on some cases, tenderly provide RPC with only 1k max block range capacity per request. This make the fetch slower, but because it's only on fork env it's alright. For the prod mode we use 100k block as max range.

### General Notes:
- I noticed that the `SupplyAssetsListMobileItem.tsx` file wasn't used because there's already a `SupplyAssetsListItemMobile` component in the `SupplyAssetsListItem.tsx` file.
I proposed to either remove the `SupplyAssetsListMobileItem.tsx` file or remove the `SupplyAssetsListItemMobile` include in the `SupplyAssetsListItem.tsx` file and use the one in `SupplyAssetsListMobileItem.tsx`

Let me know :))

---

## Author Checklist

- [ ]  The `Ready for Dev Review` label has been added

## Reviewer Checklist


src/services/UnderlyingYieldService.ts
```diff
@@ -0,0 +1,376 @@
...
+          { indexed: false, name: 'postTotalShares', type: 'uint256' },
+          { indexed: false, name: 'postTotalEther', type: 'uint256' },
+          { indexed: false, name: 'sharesMintedAsFees', type: 'uint256' },
+        ],
+        name: 'TokenRebased',
+        type: 'event',
+      },
+    ];
+
+    const contract = new Contract('0xae7ab96520DE3A18E5e111B5EaAb095312D7fE84', abi); // stETH token
```
foodaka, line 115: @MartinGbz can you use the address book

src/services/UnderlyingYieldService.ts
```diff
@@ -0,0 +1,376 @@
...
+        inputs: [
+          { indexed: true, internalType: 'address', name: 'oracle', type: 'address' },
+          { indexed: false, internalType: 'uint256', name: 'newExchangeRate', type: 'uint256' },
+        ],
+        name: 'ExchangeRateUpdated',
+        type: 'event',
+      },
+    ];
+
+    const contract = new Contract('0x9b37180d847B27ADC13C2277299045C1237Ae281', abi); // cbETH Oracle
```
foodaka, line 278: all contract addresses should be pulled from the aave-address-book. If they are not they should be added https://github.com/bgd-labs/aave-address-book

MartinGbz: Even for "tricky" contracts like RocketNetworkBalances (`0x6Cc65bF618F55ce2433f9D8d827Fc44117D81399`)? (contract that has no relation with Aave?

MartinGbz: I've just push a [PR](https://github.com/bgd-labs/aave-address-book/pull/481)
I'll tag you once the modifications will be done on this PR

sakulstra: Chiming in here as a maintainer of address-book.

We don't really want to list these addresses on address-book. The reason is that address book first and foremost is a library for usage in solidity:
> This repository contains an up-to-date registry of all addresses of the Aave ecosystem's smart contracts, for its usage in Solidity codebases.

and the js export is a low effort gimmick we added at a later point to also expose these addresses in js. This by no means implies though, that we'll add any address that might be useful in js on the address-book as this always comes at a cost for the solidity users of the library.

---

This out of the way, i had a quick look at what is done here and want to share my 2ct:

1. I think what @MartinGbz did here is quite valuable and could be useful for other ecosystem projects. I'd like to encourage you to develop this as an independent & modular package so you can rely on viem, multicall etc. We'd be happy to help & contribute.
2. probably it would be better to use `getLogs` with a topics filter over `queryFilter`. The reason is that queryFilter is not as widely supported and buggy on some node implementations.
3. For use in an ui (like this one) I think it's quite wasteful to actually do what is done here. Every user will need to do multiple api calls, fetch events over time and do calculations etc. Offloading unnecessary work to the client comes as a big cost. I would highly encourage the maintainers of this app to put 1) behind some cached api so users have to do a single request and not calculate anything clientside.

JoaquinBattilana:
This is being done right now with merit

PR #2226, opened 2024-10-23
feat: add metamask sdk

EdouardBougon:
## MetaMask SDK integration

This PR adds the connection to MetaMask using the MetaMask SDK.
We have developed a specific Web3-React connector.

The SDK must be used for both connecting via the extension and on mobile.


grothem:
Thanks for the PR @EdouardBougon 

What are the advantages to using this sdk vs. the injected connector that we're currently using.

PR #2365, opened 2025-02-17
✨Add new Ethena Rewards tooltip [skip cypress]

MartinGbz:
## General Changes

This PR adds a new "Ethena Rewards" incentive card that showcases the number of points earned by holding an eligible asset of Ethena (ie: Core USDe/sUSDe and Prime sUSDe)

## Developer Notes

/

---

## Reviewer Checklist

- [x]  If there are new environment variables being added, they have been added to the `.env.example` file as well as the pertinant `.github/actions/*` files
- [x]  There are no CI changes, or they have been approved by the DevOps and Engineering team(s)


src/hooks/useEthenaIncentives.ts
```diff
@@ -0,0 +1,18 @@
+import { AaveV3Ethereum, AaveV3EthereumLido } from '@bgd-labs/aave-address-book';
+
+const getEthenaData = (assetAddress: string): number | undefined =>
+  ETHENA_DATA_MAP.get(assetAddress);
+
+const ETHENA_DATA_MAP: Map<string, number> = new Map([
+  [AaveV3Ethereum.ASSETS.USDe.A_TOKEN, 25],
+  [AaveV3Ethereum.ASSETS.sUSDe.A_TOKEN, 5],
+  [AaveV3EthereumLido.ASSETS.sUSDe.A_TOKEN, 5],
+]);
```
foodaka, line 10: @MartinGbz this is hardcoded right? so we need a reminder when they expire basically?

MartinGbz: Yes It's hardcoded, this number didn't changed for more than 6 months now, that's why an API is a bit overkill here.
I don't think these incentives are intended to be expire, but if you want I can ask Ethena team to have a better vision on this

foodaka: got it thanks for the info

MartinGbz: I asked Ethena, the current season (3) wil end at 23/03/2025 (source: Q&A: https://app.ethena.fi/join)
But for the new season, this will need to be discussed again, so no info on ths after the 23/03

PR #2371, opened 2025-02-20
fix: zksync gas estimation from safe account

grothem:
## General Changes

- Adds in new state for if connected account is a smart contract
- On gas estimation on zk sync and for smart contract accounts, always return the hardcoded default value, since gas estimation will fail, see https://github.com/zkSync-Community-Hub/zksync-developers/discussions/144 for more details

TODO
- [x] https://github.com/aave/aave-utilities/pull/623
---

## Reviewer Checklist


src/libs/web3-data-provider/Web3Provider.tsx
```diff
@@ -179,6 +183,21 @@ export const Web3ContextProvider: React.FC<{ children: ReactElement }> = ({ chil
     }
   }, [readOnlyModeAddress, setAccount]);
 
+  console.log('account', account);
```
foodaka, line 186: we can remove the logs

foodaka:
It's failing to withdrawl ETH with the safe on zksync. I think we need to check here https://github.com/aave/interface/blob/main/src/helpers/useTransactionHandler.tsx#L407-L419 to set a default gas on zksync? `setGasLimit(BigNumber.from(230000).toString() || '');` here will fix it @grothem

foodaka (dismissed):

foodaka (approved):

JoaquinBattilana (approved):

PR #2380, opened 2025-03-02
feat: add Ether.fi loyalty program

NandyBa:
## General Changes

- Add the Ether.fi Loyalty program for weETH and allow other assets and protocol actions to be rewarded.

## Developer Notes

Note: we didn't use the same syntaxe that Ethena's Data Map before weETH is not launch yet Prime (Lido) Instance

Ether.fi confirm that incentive is already live

---

## Reviewer Checklist


src/components/incentives/EtherfiIncentivesTooltipContent.tsx
```diff
@@ -0,0 +1,30 @@
+import { Trans } from '@lingui/macro';
+import { Box } from '@mui/material';
+
+import { Link } from '../primitives/Link';
+
+export const EtherFiAirdropTooltipContent = ({ multiplier }: { multiplier: number }) => {
+  return (
+    <Box>
+      <Trans>
+        {`This asset is eligible for the Ether.fi Loyalty program with a `}
```
grothem, line 10: Could you update the text here so it's consistent with how the Sonic and Ethen ones are worded?

So something like:

```
This asset is eligible for 3x Ether.fi Loyalty Points
```

NandyBa: It's a boost on a their loyalty program. Hold only weETH make you already earn point. This incentive boost the number of points earn per token

grothem (approved):

foodaka (approved):

PR #2428, opened 2025-04-14
feat: add gnosis eure campaign

MartinGbz:
We lack the forum link for now, I'll add it as soon as we posted it


MartinGbz:
The campaign is live:

Capture d’écran 2025-04-15 à 11 03 06

foodaka (approved):

PR #2454, opened 2025-05-01
feat: added incentives stS [skip cypress]

grothem:


NandyBa:
This is already handled by #2456

PR #2484, opened 2025-05-30
feat: batch wallet capability

defispartan:
## General Changes

Add [EIP-5792](https://eip.tools/eip/5792) `atomic` wallet capability (batch transactions) for `supply`.

Add wallet capability flag to global store. If available, display batch transaction as available method and use EIP-5792 `sendCalls` RPC method to send batch transaction request to wallet. 

Currently only implemented for `supply`, the transaction method helpers are standalone components that can be re-used for different actions requiring token approval (`repay`, `stake`, `swap`. etc.).


lightclient:
Hi, I was curious when this might land?

apoorvlathey:
checking in after 1yr+, why hasn't this PR merged yet?

PR #2500, opened 2025-06-11
using family api for fetching usd prices in swaps

mgrabina:
Closes AAVE-449


src/services/FamilyPricesService.ts
```diff
@@ -0,0 +1,111 @@
...
+
+type FamilyPricesResponse = {
+  prices: {
+    [tokenId: string]: {
+      usd: number;
+    };
+  };
+};
+export class FamilyPricesService {
+  private proxyUrl = '/api/family-prices-proxy';
```
absis: I suggest to remove family name endpoint to just prices-proxy.

pages/api/family-prices-proxy.ts
```diff
@@ -0,0 +1,79 @@
+import { NextApiRequest, NextApiResponse } from 'next';
+
+const FAMILY_API_URL = 'https://experimental.staging.family.co/v3/latest-prices';
```
absis: I guess we need to use the production endpoint

absis: Instead of having it here hardcoded, we can use an environment variable. Easy to manage staging/production endpoints

.env.development
```diff
@@ -9,3 +9,5 @@ NEXT_PUBLIC_ENABLE_STAKING=true
 NEXT_PUBLIC_ENABLE_GOVERNANCE=true
 NEXT_PUBLIC_API_BASEURL=https://aave-api-v2.aave.com
 NEXT_PUBLIC_SUBGRAPH_API_KEY=
+FAMILY_API_KEY=
+FAMILY_API_URL=
```
mgrabina, line 13: @absis can u add this one to prod env?

absis: done

grothem (approved):

JoaquinBattilana (approved):

JoaquinBattilana (approved):

PR #2516, opened 2025-06-25
Implement a new Merkl incentives tooltip

MartinGbz:
## General Changes

This PR adds a new Merkl incentives tooltip: it allows to display all Aave incentives created on Merkl.
We will need this, because a Ethena USDtb incentive campaign will start soon, and Merkl manage the implementation.

A `whtelistedRewardTokens` has been created to only display incentives that are underlying or A tokens already implemented on Aave. This avoid displaying a random campaign with a random reward token that someone distribute. Even if reward token are already whitelisted on Merkl side, but this act as a double security.

Also we added the aUSDtb to the symbolMap in order to have the icon well displayed for the incentives.

This PR remove the ZKSync Ignite tooltip which has 99% the same logic than Merkl tooltip, because ZkSync Ignite was built by Merkl. It has to be removed because otherwise we would have a double tooltip display on ZKSync incentives case. Indeed the only way to differentiate a ZKSync Ignite incentives from a ""normal"" Merkl incentives was just to filter Merkl incentives by the ZKSync chain. If they decide to relaunch the program, we'll then ask the Merkl team to come up with a way of differentiating these incentives. In the meantime, the better is to just remove it (in any case now, every Merkl incentives, ZKSync ignite or ont, will be displayed).

## Developer Notes

In order to test the implementation you can do this update in the `useMerklincentives.ts` file:

```
// if (opportunity.status !== OpportunityStatus.LIVE) {
//   return null;
// }

if (opportunity.apr <= 0) {
  // return null;
  opportunity.apr = 99;
}
```

This will hardcode a 99% APR on all asset that are previously incentivized on Merkl

Here's the result:

Capture d’écran 2025-06-25 à 15 50 52

---

## Reviewer Checklist


package.json
```diff
@@ -154,5 +154,6 @@
     "budget": null,
     "budgetPercentIncreaseRed": 20,
     "showDetails": true
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
```
NandyBa:
```suggestion
  }
```

src/hooks/useMerklIncentives.ts
```diff
@@ -28,31 +44,59 @@ type MerklOpportunity = {
...
+}) => {
+  return Object.entries(assets).flatMap(([, asset]) => [
+    asset.UNDERLYING.toLowerCase(),
+    asset.A_TOKEN.toLowerCase(),
+  ]);
+};
+
+const whitelistedRewardTokens = allAaveAssets.flatMap((assets) => getUnderlyingAndAToken(assets));
+
+const url = 'https://api.merkl.xyz/v4/opportunities?mainProtocolId=aave'; // Merkl API
```
foodaka:
```suggestion
const MERKL_ENDPOINT = 'https://api.merkl.xyz/v4/opportunities?mainProtocolId=aave'; // Merkl API
```

src/hooks/useMerklIncentives.ts
```diff
@@ -79,15 +123,11 @@ export const useZkSyncIgniteIncentives = ({
...
   return useQuery({
     queryFn: async () => {
-      if (market === CustomMarket.proto_zksync_v3) {
-        const response = await fetch(url);
-        const merklOpportunities: MerklOpportunity[] = await response.json();
-        return merklOpportunities;
-      } else {
-        return [];
-      }
+      const response = await fetch(`${url}`);
```
foodaka:
```suggestion
      const response = await fetch(`${MERKL_ENDPOINT}`);
```

foodaka (approved):

PR #2527, opened 2025-07-03
Swaps flow improvements

mgrabina:
This PR includes a bunch of UX improvements, and a few fixes and safe checks for border cases, mostly recommended by our support and CoW's QAs.

Some of the changes are:
- Slippage Selector: 
   - showing 'Auto' or 'Custom' slippage title
   - improved logic for default options, adding Auto as well
   - native tokens minimum per chain improved ux
- blocking max input when no need to use it
- show exipred past txs as expired not cancelled
- show user rejected tx message
- refetching quote each 30s
- fix responsiveness for swaps history in mobile
- show full amounts in tooltips (usefull for small ones)
- more precise eth leftover for layer 2s in eth-flows
- showing spot values for easy comparison with other swap protocols
- improved surplus text for small amounts
- eth-flow not allowed in non-Safe SC wallets
- special revoke+approval flow for second higher approvals for USDT in mainnet
- improved cow orders metadata for analytics
- nit widget title when no asset is selected
- slippage cap 
- adds cow to tokens list


src/components/transactions/Switch/BaseSwitchModalContent.tsx
```diff
@@ -667,6 +690,17 @@ export const BaseSwitchModalContent = ({
                 </Warning>
               )}
 
+              {showUSDTResetWarning && (
```
foodaka, line 746: wondering if we should guide them more here. Like send them somewhere in open tab to do this

foodaka: ah i see we are doing that down below?

mgrabina: yeah simpler: we just use same approval button, and after reset the warning "required additonal tx" is removed and same flow as usual

foodaka (approved):

foodaka:
@MartinGbz this one reported - can we check the other native tokens as well

Screenshot 2025-07-08 at 11 52 34

mgrabina:
> @MartinGbz this one reported - can we check the other native tokens as well
> 
> Screenshot 2025-07-08 at 11 52 34

done, the issue was just for avax as it was coming in lowercase from cow's sdk, wrapping is not supported by cow solvers and we will need to do a separate flow (no fees)

foodaka (approved):

PR #2586, opened 2025-08-14
Feat/2576 add underlying in umbrella

AGMASO:
## General Changes

- Added a new column in both Desktop and Mobile views to display `Shares`, representing the amount of `shares/stkTokens ` the user holds.

- Renamed `AmountStakedItem.tsx` to `AmountStakedUnderlyingItem.tsx` and updated the logic to use `stakeTokenRedeemableAmount` for displaying the amount of underlying asset tokens (e.g., USDC, WETH...) the user has staked in the protocol.

## Developer Notes
- I noticed that `stakeTokenRedeemableAmount` is not functioning as expected.

It currently returns the amount of `stkToken` that can be redeemed, not the result from `previewRedeem()` (i.e., the equivalent amount of underlying asset tokens for the given shares).


AGMASO:
Closing this draft PR in favor of a new one with an updated description and final changes.

PR #2593, opened 2025-08-20
feat: added dynamic assets categories using Coingecko API

AGMASO:
## General Changes

- Added dynamic categories and asset filters on the Market page using the Coingecko API. We can now filter by: All, Stablecoins, or ETH-Correlated.

- Improved mobile view with layout modifications.

- Added dynamic categories and asset filters also in Dashboard page.

## Developers Notes:

- NOTE: You need to add the API key for --> `COINGECKO_API_KEY=`

---

## Reviewer Checklist


AGMASO:
Added the same style as in Figma, using the ReservePage timeframe buttons as a reference to ensure consistency across the app.

src/modules/markets/utils/useCoinGeckoStablecoinCat.ts
```diff
@@ -0,0 +1,71 @@
+import { useEffect, useState } from 'react';
+
+import { STABLECOINS_SYMBOLS_FALLBACK } from './assetCategories';
+
+const CG_ENDPOINT = 'https://pro-api.coingecko.com/api/v3/coins/markets';
+const HEADERS: HeadersInit = {
+  accept: 'application/json',
+  'x-cg-pro-api-key': process.env.NEXT_PUBLIC_CG_API_KEY ?? '',
```
mgrabina: let's put the api key behind a proxy so we hide it from external abuse

src/modules/markets/utils/useCoinGeckoStablecoinCat.ts
```diff
@@ -0,0 +1,71 @@
...
+  const [loading, setLoading] = useState(false);
+  const [error, setError] = useState<Error | null>(null);
+
+  useEffect(() => {
+    const loadStablecoins = async () => {
+      setLoading(true);
+      setError(null);
+
+      try {
+        // Toda la lógica de fetch integrada aquí
```
mgrabina: english if a comment if necessary (this one we can delete yo creo)

src/modules/markets/utils/useCoinGeckoStablecoinCat.ts
```diff
@@ -0,0 +1,71 @@
...
+        const [response1, response2] = await Promise.all([
+          fetch(`${CG_ENDPOINT}?vs_currency=usd&category=stablecoins&per_page=250&page=1`, {
+            method: 'GET',
+            headers: HEADERS,
+          }),
+          fetch(`${CG_ENDPOINT}?vs_currency=usd&category=stablecoins&per_page=250&page=2`, {
+            method: 'GET',
+            headers: HEADERS,
+          }),
```
mgrabina: - how we make sure 2 pages will be enough
- maybe we can bring more items per page so more efficient
- can we ask for a specific list? https://docs.coingecko.com/reference/coins-id not sure

src/modules/markets/utils/useCoinGeckoStablecoinCat.ts
```diff
@@ -0,0 +1,71 @@
+import { useEffect, useState } from 'react';
+
+import { STABLECOINS_SYMBOLS_FALLBACK } from './assetCategories';
+
+const CG_ENDPOINT = 'https://pro-api.coingecko.com/api/v3/coins/markets';
```
mgrabina: how we define the chain id? so which chain ids are supported? e.g. i can't see it loading on zkSync not sure if bc it's not supported by coingecko something to double check

src/modules/markets/utils/useCoinGeckoStablecoinCat.ts
```diff
@@ -0,0 +1,71 @@
...
+        setStablecoinSymbols(uniqueSymbols);
+      } catch (err) {
+        setError(err as Error);
+        console.error('Error loading stablecoins from CoinGecko, using fallback:', err);
+      } finally {
+        setLoading(false);
+      }
+    };
+
+    loadStablecoins();
```
mgrabina: would be possible to merge all 'use' hooks and have a simpler useAssetCategories? e.g. this may grow in the future too

src/modules/markets/utils/useCoinGeckoStablecoinCat.ts
AGMASO: - In the CoinGecko API, we can’t set more than 250 items per page.

- The full list of the stablecoins category has 356 items, so two pages are enough.

- We could request specific tokens, but in this case I think the goal is to fetch the entire pre-made category so we don’t have to maintain the list ourselves.

src/modules/markets/utils/useCoinGeckoStablecoinCat.ts
AGMASO: Sorry!

src/modules/markets/utils/useCoinGeckoStablecoinCat.ts
AGMASO: I will do it

src/modules/markets/utils/useCoinGeckoStablecoinCat.ts
AGMASO: To retrieve coins from a category in CoinGecko, we can’t specify a chain ID. The API returns the general tokens (without bridge variations like USDC.e for USDC) as well as chain-specific tokens when they exist. That’s why we can use the category endpoint directly, without worrying about chain IDs.

src/modules/markets/utils/useCoinGeckoCategories.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export function useCoingeckoCategories() {
+  return useQuery({
+    queryFn: async () => {
+      const response = await fetch('/api/coingecko-categories');
+      if (!response.ok) throw new Error('Failed to fetch CoinGecko categories');
+      return await response.json();
+    },
+    queryKey: ['coingecko-categories'],
+    staleTime: 1000 * 60 * 60 * 24,
```
mgrabina: maybe we can even cache the response for a week?

pages/api/coingecko-categories.ts
```diff
@@ -0,0 +1,144 @@
...
+      }),
+      fetch(`${CG_ENDPOINT}?vs_currency=usd&category=ether-fi-ecosystem&per_page=250&page=1`, {
+        method: 'GET',
+        headers: HEADERS,
+      }),
+      fetch(`${CG_ENDPOINT}?vs_currency=usd&category=liquid-staking-tokens&per_page=250&page=1`, {
+        method: 'GET',
+        headers: HEADERS,
+      }),
+    ]);
```
mgrabina: we can make eth calls in parallel to stable to optimize

pages/api/coingecko-categories.ts
```diff
@@ -0,0 +1,144 @@
...
+  } else if (origin && isOriginAllowed(origin)) {
+    res.setHeader('Access-Control-Allow-Origin', origin);
+  }
+
+  res.setHeader('Access-Control-Allow-Methods', 'GET, OPTIONS');
+  res.setHeader('Access-Control-Allow-Headers', 'Content-Type');
+
+  if (req.method === 'OPTIONS') {
+    return res.status(200).end();
+  }
```
mgrabina, line 39: really nice!

src/modules/markets/MarketAssetsListContainer.tsx
```diff
@@ -63,6 +71,15 @@ export const MarketAssetsListContainer = () => {
...
     })
+    // Filter by category
+    .filter((res) =>
+      isAssetInCategoryDynamic(
+        res.symbol,
+        selectedCategory,
+        stablecoinSymbols,
+        ethCorrelatedSymbols
+      )
+    )
```
mgrabina: nice! so clean

src/modules/markets/utils/assetCategories.ts
```diff
@@ -0,0 +1,143 @@
+export enum AssetCategory {
+  ALL = 'all',
+  STABLECOINS = 'stablecoins',
+  ETH_CORRELATED = 'eth_correlated',
+}
+
+export const STABLECOINS_SYMBOLS_FALLBACK = [
```
mgrabina: interesting. so i imagine the effect if the request takes time and I've selected a category will see the list to update .e.g. adding more items because coingecko has more than the fallback, wondering if it makes sense to have a fallback or just disable the picker until we have data. Plus, we can cache the data in the localStorage for a long time, since categories wont change frequently.

mgrabina (commented):
AMAZING progress since the past iteration! would like to have others reviews as well as we have some design desicions to do!

src/modules/markets/utils/useCoinGeckoCategories.ts
AGMASO: done

pages/api/coingecko-categories.ts
AGMASO: done

src/modules/markets/utils/assetCategories.ts
AGMASO: Understood. I was thinking it would be a good idea to always render something for the user, for example, if the CoinGecko API is down. That said, I agree the other approach makes sense as well. Let’s make a final decision with the rest of the team, and I can update it, no problem.

PR #2597, opened 2025-08-21
feat: add plain support integration to the support section

AGMASO:
## General Changes

- add Plain support platform integration to the `Get Support` section

## Developer Notes

- Requires adding `API_KEY` to the production `.env` file

---

## Reviewer Checklist


mgrabina (commented):
image

we can improve a bit the look and feel to match overall app:
- the "let us know.." no margin and using white/subtitle, we should always reuse theme.palette stuff
- the inputs with same padding and no title as the search bars? not sure if we have any other form in the site
- remove the points from the texts and double check copys with Kolten

image

some bottom padding to make it more symmetric? also Ok,Close not closing to me

didnt receive a confirmation email not sure if intended but let's make sure it works and we receive them 👍🏼

pages/api/support-create-ticket.ts
```diff
@@ -0,0 +1,157 @@
...
+
+    if (allowedOrigins.includes(origin)) return true;
+
+    // Match any subdomain ending with avaraxyz.vercel.app for deployment urls
+    const allowedPatterns = [/^https:\/\/.*avaraxyz\.vercel\.app$/];
+
+    return allowedPatterns.some((pattern) => pattern.test(origin));
+  };
+
+  if (process.env.CORS_DOMAINS_ALLOWED === 'true') {
```
foodaka: this is never set CORS_DOMAINS_ALLOWED

AGMASO: deleted

foodaka (approved):

foodaka (approved):

package.json
```diff
@@ -53,6 +53,7 @@
     "@safe-global/safe-apps-provider": "^0.18.4",
     "@safe-global/safe-apps-sdk": "^9.1.0",
     "@tanstack/react-query": "^5.62.8",
+    "@team-plain/typescript-sdk": "^5.10.3",
```
foodaka: @AGMASO can we just have a quick check if this is necessary? Can we check if we can just call their graph ql endpoint directly

pages/api/support-create-ticket.ts
```diff
@@ -0,0 +1,155 @@
+import {
+  ComponentDividerSpacingSize,
+  ComponentSpacerSize,
+  ComponentTextColor,
+  ComponentTextSize,
+  PlainClient,
```
foodaka: for example this stuff is just UI

package.json
AGMASO: removed

pages/api/support-create-ticket.ts
AGMASO: done

PR #2616, opened 2025-09-02
feat: tooltip updated with self incentives

AGMASO:
## General Changes

- Added Self campaign–specific changes to the incentives tooltip (split into Merit Incentive and Self Incentive, updated copy).

- Updated Total APY logic when Self applies: Protocol APY + Merit APY + Self APY.

## Developer Notes

- Uses keys: `celo-supply-<asset>` (Merit) and `self-<action>` (Self).

- If Self APR is null/0, the Self Incentive row is hidden and Total APY excludes it.

---

## Reviewer Checklist


NandyBa:
Result seems good to be be for Celo Self campaign
It will require an other PR for displaying at the same time Merit supply wETH & Merit supply multiple and borrow USDT

src/components/incentives/IncentivesCard.tsx
```diff
@@ -75,6 +75,8 @@ export const IncentivesCard = ({
   });
 
   const meritIncentivesAPR = meritIncentives?.breakdown?.meritIncentivesAPR || 0;
+  const selfAPY = ENABLE_SELF_CAMPAIGN ? meritIncentives?.variants?.selfAPY ?? 0 : 0;
```
foodaka, line 80: i think this is a one off so we can leave a note to remove it after these incentives are finished

AGMASO: done

foodaka (approved):

grothem (approved):

PR #2627, opened 2025-09-08
Feat/Swap Modals UX improvements related to Slippage selector and bypass low HF warning

AGMASO:
## General Changes

- Closes ALE-7 :  Enabled users to approve and bypass the low HF warning on collateral swaps: 

 If HF is between 1.05 and 1.01, an alert is shown along with a checkbox the user must confirm.

 If HF is below 1.01, a warning is shown and the user is not allowed to proceed with the swap.

- Closes ALE-10 :  Disabled the slippage selector until the first suggested slippage is available, and ensured that a custom slippage value remains selected even if a new suggestion comes from a fresh quote.

- Closes ALE-15 : Removed the word "Auto" from the slippage selector when the Paraswap flow is used.

- Set the Paraswap default slippage to 0.2% when assets are not correlated (e.g., WBTC ↔ USDT), while keeping 0.1% for correlated assets.

- Closes ALE-8 :  Increased the debouncing time from 300ms to 1500ms. With 300ms, a quote is triggered on every number change — it technically works correctly, but the effect is that if a user changes a digit and pauses (before typing the next one), a new quote request is already executed. By increasing to 1500ms, we achieve the intended debouncing behavior, reducing the number of quote calls. Since a Cow/Paraswap calculation takes a few seconds, this provides a smoother and more user-friendly experience.

- Closes ALE-14 : Persisted the user’s token selection if the swap modal is accidentally closed. Currently, implemented only for `BaseSwitchModal`, using `sessionStorage` while differentiating between different ModalTypes.

## Developer Notes

- Currently using a hardcoded mapping of correlated assets for Paraswap and CoW. A potential improvement would be to replace the hardcoded list with the CoinGecko categories hook for a more dynamic approach.

- Components such as `WithdrawAndSwitchModal` or `DebtSwitch` do not use the `BaseSwitchModal`, so the behavior of preserving the token selection is not applied there.
---

## Reviewer Checklist


mgrabina:
@AGMASO if you want Linear can track automatically your issues states from PRs by adding "Closes ISSUE_CODE, ISSUE_CODE_2.."

src/components/transactions/Switch/CollateralSwap/CollateralSwapModal.tsx
```diff
@@ -159,19 +159,15 @@ export const CollateralSwapModal = () => {
...
       ? tokensFrom[0]
       : undefined);
-  const defaultOutputToken =
-    tokensTo.length > 0
-      ? tokensTo.filter((token) => token.address !== defaultInputToken?.address)[0]
-      : undefined;
 
   return (
+    // Only forcedDefaultInputToken forced
     <BaseSwitchModal
```
mgrabina, line 174: what was the reason?

src/components/transactions/Switch/assetCorrelation.helpers.ts
mgrabina, line 1: nice idea to separate!

src/components/transactions/Switch/BaseSwitchModalContent.tsx
```diff
@@ -340,12 +355,58 @@ export const BaseSwitchModalContent = ({
...
+      return parsedTokenSelection;
+    } catch (e) {
+      return null;
+    }
+  };
+  const [selectedInputToken, setSelectedInputToken] = useState<TokenInfoWithBalance>(() => {
+    if (forcedDefaultInputToken) return forcedDefaultInputToken;
+
+    const saved = loadTokenSelection();
+    return saved?.inputToken || defaultInputToken;
```
mgrabina, line 413: interesting how this works when CollateralModal sets a default input/ouput that the user already saw and selected other, while re opening will see the forced one by the modal right? maybe we should change namings for clarity like "default token" and ensure user selection is always prioritized?

src/components/transactions/Switch/BaseSwitchModalContent.tsx
```diff
@@ -433,6 +494,16 @@ export const BaseSwitchModalContent = ({
...
+  const getParaswapSlippage = (inputSymbol: string, outputSymbol: string): string => {
+    const inputGroup = getAssetGroup(inputSymbol);
+    const outputGroup = getAssetGroup(outputSymbol);
+
+    if (!inputGroup || !outputGroup || inputGroup === 'unknown' || outputGroup === 'unknown') {
+      return '0.20';
+    }
+    return inputGroup === outputGroup ? '0.10' : '0.20';
+  };
+
```
mgrabina: niice! thinking that slippage logic is growing a lot maybe we should separate it another file?

mgrabina: also this condition may be more clear viceversa: if equal .0.1, otherwise just 0.2 no more conditions

src/components/transactions/Switch/BaseSwitchModalContent.tsx
```diff
@@ -532,7 +625,16 @@ export const BaseSwitchModalContent = ({
...
 
-    return hfNumber.lt(LIQUIDATION_SAFETY_THRESHOLD);
+    return hfNumber.lt(LIQUIDATION_SAFETY_THRESHOLD) && hfNumber.gte(LIQUIDATION_DANGER_THRESHOLD);
+  }, [hfAfterSwap]);
+  const isLiquidatable = useMemo(() => {
+    if (!hfAfterSwap) return false;
+
+    const hfNumber = new BigNumber(hfAfterSwap);
+
+    if (hfNumber.lt(0)) return false;
```
mgrabina, line 636: good catch

src/components/transactions/Switch/BaseSwitchModalContent.tsx
```diff
@@ -960,16 +1072,66 @@ export const BaseSwitchModalContent = ({
...
+                    alignItems: 'center',
+                  }}
+                >
                   <Typography variant="caption">
                     <Trans>
-                      Low health factor after swap. Please select a different asset or lower the
-                      amount.
+                      Your health factor after this swap will be critically low and you will be
+                      liquidated. Please choose a different asset or reduce the swap amount to stay
+                      safe.
```
mgrabina: not necessarily liquidated, would write something like, this is a risky swap as your HF will be near liquidiation threshold, proceed with extreme caution / something like that better written by ai

src/components/transactions/Switch/BaseSwitchModalContent.tsx
```diff
@@ -960,16 +1072,66 @@ export const BaseSwitchModalContent = ({
...
+                    mt: 2,
+                    mb: 2,
+                    display: 'flex',
+                    flexDirection: 'column',
+                    alignItems: 'center',
+                  }}
+                >
+                  <Typography variant="caption">
+                    <Trans>
+                      Low health factor after swap. This may put you at risk of liquidation.
```
mgrabina: Also less needed lquidation wording in this one that has more margin

src/components/transactions/Switch/BaseSwitchModalContent.tsx
mgrabina: oh okay we have i understan the risk thats nice

src/components/transactions/Switch/SwitchSlippageSelector.tsx
```diff
@@ -30,7 +31,9 @@ const defaultSlippageOptions = (suggested?: string) => {
   }
 
   const suggestedNumber = Number(suggested);
-
+  if (suggestedNumber <= 0.1) {
+    return ['0.20', '0.50', 'Auto'];
+  }
```
mgrabina, line 36: is this intended? doesnt make much sense right?

src/components/transactions/Switch/SwitchSlippageSelector.tsx
```diff
@@ -66,17 +70,27 @@ export const SwitchSlippageSelector = ({
...
+  }, [slippage, userHasSetCustomSlippage, previousSlippage, setSlippage]);
+
+  //Update slippage to suggested if user has not set a custom slippage
+  useEffect(() => {
+    if (suggestedSlippage && !userHasSetCustomSlippage && !isCustomSlippage) {
+      setSlippage(suggestedSlippage);
+      setPreviousSlippage(slippage);
     }
-  }, [slippage]);
+  }, [suggestedSlippage, userHasSetCustomSlippage, isCustomSlippage, setSlippage]);
```
mgrabina: can we merge useEffects?

src/components/transactions/Switch/CollateralSwap/CollateralSwapModal.tsx
mgrabina: image
side effect getting same default input/output

mgrabina (commented):
amazing work!

src/components/transactions/Switch/CollateralSwap/CollateralSwapModal.tsx
mgrabina: @AGMASO what about this one

src/components/transactions/Switch/SwitchSlippageSelector.tsx
mgrabina: @AGMASO usually the user would select a lower one than suggested, that's why

src/components/transactions/Switch/CollateralSwap/CollateralSwapModal.tsx
```diff
@@ -159,19 +159,28 @@ export const CollateralSwapModal = () => {
...
     tokensTo.length > 0
-      ? tokensTo.filter((token) => token.address !== defaultInputToken?.address)[0]
+      ? tokensTo.filter(
+          (token) =>
+            token.address !== defaultInputToken?.address &&
+            token.symbol !== defaultInputToken?.symbol
+        )[0]
       : undefined;
+  console.log('Default Input Token: ', defaultInputToken);
+  console.log('Default Output Token: ', defaultOutputToken);
```
mgrabina: lets remove any logs before merge

mgrabina (approved):

mgrabina (approved):

src/components/transactions/Switch/CollateralSwap/CollateralSwapModal.tsx
foodaka:
```suggestion
```

foodaka:
@AGMASO can you rebase main

mgrabina (approved):

PR #2631, opened 2025-09-10
feat: add Add to wallet button, show USD amounts and fix small-number rounding

AGMASO:
## General Changes

- Improved the UX of SwitchTxSuccessView.tsx by adding USD values as a subheader.

- Added an “Add to wallet” button to integrate the received token into the user’s wallet.

- Now we don’t round small numbers. If the number is smaller than 4 decimals, we will show <0.00001.

## Developer Notes

* Regarding the **“Add to wallet”** button, there is an issue we need to discuss. At the moment, the token to be added to the wallet is the underlying token. The problem is that in the case of **CollateralSwitch**, we are actually switching aTokens. This creates an inconsistency issue in other markets, because the aToken symbol will be → `a{nameOfChainAbbreviation}{tokenName}`. This causes a conflict when sending the RPC to the wallet, since the parameters we have only include the symbol of the normal aToken, not the market-specific one.

**Questions:**

1. Is it intended to add the `a{nameOfChainAbbreviation}{tokenName}`, or should this only be done in the case of underlying tokens?
2. Should we remove this button when performing a collateral swap, and leave it only for normal token swaps?

## Developer Notes

- Con respecto al boton "Add to wallet", existe un problema que debemos discutir. En estos momentos, el token a agregar en la wallet sera el underlying token, el probloema es que en el caso de CollateralSwitch, estamso cambiando aTokens, Esto generea un issue de inconsitencia cuando estamso en otros markets porque el aToken symbol sera igual a --> `a{nameOfChainAbreviatura}{tokenName}` esto hace que haya un conglicto al enviar el RPC a la wallet proque los parametros de los que contamos no se encuntra el symbol de este token si no del AToken normal sin mercado incluido
Preguntas:
1- Es intended agreagar el `a{nameOfChainAbreviatura}{tokenName}` , o solo en el caso de underlying tokens? 
2- debemos quitar este button cuando se esta haceindo un collateral swap? y dejarlo solo para normal swaps de tokens? 

---

## Reviewer Checklist

### DIFF
--- a/src/components/transactions/Switch/BaseSwitchModalContent.tsx
+++ b/src/components/transactions/Switch/BaseSwitchModalContent.tsx
@@ -200,6 +200,7 @@ export const BaseSwitchModalContent = ({
 
   const [userIsSmartContractWallet, setUserIsSmartContractWallet] = useState(false);
   const [userIsSafeWallet, setUserIsSafeWallet] = useState(false);
+
   useEffect(() => {
     try {
       if (user && connectedChainId) {
@@ -684,6 +685,15 @@ export const BaseSwitchModalContent = ({
           .multipliedBy(1 - safeSlippage)
           .decimalPlaces(switchRates.destDecimals, BigNumber.ROUND_UP)
           .toString()}
+        amountUsd={Number(switchRates.srcUSD)}
+        outAmountUSD={Number(
+          switchRates.provider === 'cowprotocol' ? switchRates.destSpotInUsd : switchRates.destUSD
+        )}
+        addToken={{
+          address: selectedOutputToken.address,
+          symbol: selectedOutputToken.symbol,
+          decimals: selectedOutputToken.decimals,
+        }}
       />
     );
   }
--- a/src/components/transactions/Switch/SwitchTxSuccessView.tsx
+++ b/src/components/transactions/Switch/SwitchTxSuccessView.tsx
@@ -1,13 +1,16 @@
 import { normalize } from '@aave/math-utils';
 import { Trans } from '@lingui/macro';
-import { Box, CircularProgress, Divider, Typography } from '@mui/material';
+import { Box, Button, CircularProgress, Divider, Typography } from '@mui/material';
 import { BigNumber } from 'ethers';
 import { useEffect, useMemo, useRef, useState } from 'react';
+import { WalletIcon } from 'src/components/icons/WalletIcon';
 import { DarkTooltip } from 'src/components/infoTooltips/DarkTooltip';
 import { FormattedNumber } from 'src/components/primitives/FormattedNumber';
-import { ExternalTokenIcon } from 'src/components/primitives/TokenIcon';
+import { Base64Token, ExternalTokenIcon, TokenIcon } from 'src/components/primitives/TokenIcon';
 import { TextWithTooltip, TextWithTooltipProps } from 'src/components/TextWithTooltip';
 import { useCowOrderToast } from 'src/hooks/useCowOrderToast';
+import { useWeb3Context } from 'src/libs/hooks/useWeb3Context';
+import { ERC20TokenType } from 'src/libs/web3-data-provider/Web3Provider';
 import { networkConfigs } from 'src/ui-config/networksConfig';
 import { parseUnits } from 'viem';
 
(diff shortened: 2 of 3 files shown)


src/components/transactions/Switch/BaseSwitchModalContent.tsx
```diff
@@ -684,6 +685,15 @@ export const BaseSwitchModalContent = ({
...
           .toString()}
+        amountUsd={Number(switchRates.srcUSD)}
+        outAmountUSD={Number(
+          switchRates.provider === 'cowprotocol' ? switchRates.destSpotInUsd : switchRates.destUSD
+        )}
+        addToken={{
+          address: selectedOutputToken.address,
+          symbol: selectedOutputToken.symbol,
+          decimals: selectedOutputToken.decimals,
+        }}
```
mgrabina, line 697: just for aTokens/collateral swaps

mgrabina:
image
feew ux nits

let's add screenshots to the PR to accelerate reviews too

src/components/transactions/Switch/BaseSwitchModalContent.tsx
AGMASO: Ok, I will keep it only for aToken/collateral swaps. In the Swap module for underlying token swaps will not be shown.

src/components/transactions/Switch/SwitchTxSuccessView.tsx
```diff
@@ -195,6 +207,17 @@ export const SwitchTxSuccessView = ({
...
+  const getATokenSymbol = (marketName: string, tokenSymbol: string) => {
+    const marketPrefix = marketName.slice(0, 3).toLowerCase();
+    return `a${marketPrefix}${tokenSymbol}`;
+  };
+  const finalTokenSymbol: string =
+    addToken?.aToken && addToken
+      ? isEthereum
+        ? `a${addToken.symbol}`
+        : getATokenSymbol(networkConfigs[chainId].name, addToken.symbol)
+      : addToken?.symbol ?? '';
```
AGMASO: @mgrabina I have fixed it so that we are showing only for aToken swaps. I also created this helper to define the correct token symbol for the aToken depending on each market. It is working on the Base market and makes adding tokens seamless, but I’m not able to test on mainnet and other markets as I don’t have funds, and using a fork is not working.

Before approving this, we need to test that the correct aToken symbol is created in different markets to avoid failures when adding the token to the wallet.

AGMASO:
@mgrabina the margins + paddings are now fixed
## Normal Swap, no collateral
Screenshot 2025-09-16 at 15 57 35 (2)
## aTokens Swap finished
Screenshot 2025-09-16 at 16 05 45 (2)
## aToken Swap In Progress
Screenshot 2025-09-16 at 17 15 20 (2)

src/components/transactions/Switch/BaseSwitchModalContent.tsx
```diff
@@ -684,9 +685,21 @@ export const BaseSwitchModalContent = ({
...
+          address: selectedOutputToken.aToken ?? selectedOutputToken.address,
+          symbol: selectedOutputToken.symbol,
+          decimals: selectedOutputToken.decimals,
+          aToken: selectedOutputToken.aToken ? true : false,
+        }}
       />
     );
   }
+  console.log('output token', `a${selectedOutputToken.symbol}`);
+  console.log('atoken token', selectedOutputToken.aToken);
```
foodaka: @AGMASO lets remove

src/components/transactions/Switch/SwitchTxSuccessView.tsx
```diff
@@ -94,7 +105,8 @@ export const SwitchTxSuccessView = ({
 
   // Market for chain id
   const networkConfig = networkConfigs[chainId].explorerLink;
-
+  const marketName = networkConfigs[chainId].name;
+  console.log('networkConfig', marketName);
```
foodaka: remove debug

src/components/transactions/Switch/SwitchTxSuccessView.tsx
mgrabina: hm i think is safer to just use ModalType instead of inferring from the token symbol that cannot be standarized

src/components/transactions/Switch/SwitchTxSuccessView.tsx
```diff
@@ -223,11 +246,10 @@ export const SwitchTxSuccessView = ({
       <Box
         sx={{
           background: 'background.default',
-          borderRadius: 2,
+          borderRadius: '8px',
```
mgrabina: is this needed? does it affect responsiveness?

mgrabina (changes requested):

PR #2641, opened 2025-09-16
fix: fix Horizon banner redirection when navigating from the sGHO page

AGMASO:
## General Changes

- fixed Horizon banner redirection when navigating from the sGHO page

## Developer Notes

- Could `currentMarket` change while we’re on the `/sGho` page? This might be an edge case. I don’t think it’s possible.
We can’t add `currentMarket` to the `useEffect` dependency array because it conflicts with the banner redirection.

---

## Reviewer Checklist

### DIFF
--- a/pages/sgho.page.tsx
+++ b/pages/sgho.page.tsx
@@ -3,6 +3,7 @@ import { AaveV3Ethereum } from '@bgd-labs/aave-address-book';
 import { Trans } from '@lingui/macro';
 import { Box, Paper, Typography, useMediaQuery, useTheme } from '@mui/material';
 import dynamic from 'next/dynamic';
+import { useRouter } from 'next/router';
 import { useEffect } from 'react';
 import { ContentContainer } from 'src/components/ContentContainer';
 import { ConnectWalletButton } from 'src/components/WalletConnection/ConnectWalletButton';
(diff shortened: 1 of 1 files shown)


pages/sgho.page.tsx
```diff
@@ -49,14 +50,16 @@ export default function SavingsGho() {
   const { data: stakeUserResult } = useUserStakeUiData(currentMarketData);
 
   const { data: stakeGeneralResult } = useGeneralStakeUiData(currentMarketData);
-
-  // Automatically switch to mainnet if not already on mainnet
+  const router = useRouter();
+  console.log('Router path:', router.pathname);
+  // Automatically switch to mainnet if not already on mainnet when entering the sGHO page
   // since sGHO only exists on Ethereum mainnet
+  // NOTE: Having currentMarket as a dependency in useEffect causes conflicts with the Horizon Banner interaction.
```
mgrabina, line 56: what exactly

pages/sgho.page.tsx
```diff
@@ -49,14 +50,16 @@ export default function SavingsGho() {
   const { data: stakeUserResult } = useUserStakeUiData(currentMarketData);
 
   const { data: stakeGeneralResult } = useGeneralStakeUiData(currentMarketData);
-
-  // Automatically switch to mainnet if not already on mainnet
+  const router = useRouter();
+  console.log('Router path:', router.pathname);
```
mgrabina: lets remove comment before merge

pages/sgho.page.tsx
AGMASO: When we click the link in the banner to go to the Horizon Markets page, the useEffect detects that `currentMarket` has changed and re-renders the component, preventing us from reaching the desired page. The solution is to remove `currentMarket` from the useEffect dependency array.

mgrabina (approved):

PR #2662, opened 2025-09-23
Refactor/transaction history with sdk, using react hook

AGMASO:
## General Changes 

* Closes AAVE-1744: Refactored history transactions using **aave-sdk**
* In this version, the React hook is used
* Fixed Csv/Json `formatTransactionData`, since the current version is broken regarding CSV creation for Cow orders
* V2 is deprecated as we discussed. Only V3 is supported.
## Developer Notes

- The reason I used the React hook directly is because I tried following the documentation to use the TypeScript action method, but it’s failing due import errors. Maybe the docs are not updated for the newer versions of the SDK?
---

## Reviewer Checklist

### DIFF
--- a/src/hooks/useTransactionHistory.tsx
+++ b/src/hooks/useTransactionHistory.tsx
@@ -38,9 +38,7 @@ const sortTransactionsByTimestampDesc = (
   a: TransactionHistoryItemUnion,
   b: TransactionHistoryItemUnion
 ) => {
-  const aTime = '__typename' in a ? new Date(a.timestamp).getTime() : a.timestamp * 1000;
-  const bTime = '__typename' in b ? new Date(b.timestamp).getTime() : b.timestamp * 1000;
-  return bTime - aTime;
+  return new Date(b.timestamp).getTime() - new Date(a.timestamp).getTime();
 };
 
 export const applyTxHistoryFilters = ({
@@ -343,7 +341,7 @@ export const useTransactionHistory = ({ isFilterActive }: { isFilterActive: bool
         return {
           action: srcToken.isAToken ? 'CowCollateralSwap' : 'CowSwap',
           id: order.uid,
-          timestamp: Math.floor(new Date(order.creationDate).getTime() / 1000),
+          timestamp: new Date(order.creationDate).toISOString(),
           underlyingSrcToken: {
             underlyingAsset: srcToken.address,
             name: srcToken.name,
--- a/src/modules/history/types.ts
+++ b/src/modules/history/types.ts
@@ -12,7 +12,7 @@ import { OrderStatus } from '@cowprotocol/cow-sdk';
 export type TransactionHistoryItem<T = unknown> = {
   id: string;
   action: string;
-  timestamp: number;
+  timestamp: string;
 } & T;
 
 export type ReserveSubset = {
@@ -30,7 +30,6 @@ export type CowSwapSubset = {
   srcAmount: string;
   destAmount: string;
   status: OrderStatus;
-  timestamp: number;
   orderId: string;
   chainId: number;
 };
--- a/src/modules/history/TransactionMobileRowItem.tsx
+++ b/src/modules/history/TransactionMobileRowItem.tsx
@@ -8,12 +8,7 @@ import { GENERAL } from 'src/utils/events';
 import { useShallow } from 'zustand/shallow';
 
 import { ActionDetails, ActionTextMap } from './actions/ActionDetails';
-import {
-  getExplorerLink,
-  getTransactionAction,
-  getTransactionTimestamp,
-  unixTimestampToFormattedTime,
-} from './helpers';
+import { getExplorerLink, getTransactionAction, unixTimestampToFormattedTime } from './helpers';
 import { TransactionHistoryItemUnion } from './types';
 
 function ActionTitle({ action }: { action: string }) {
@@ -36,7 +31,7 @@ function TransactionMobileRowItem({ transaction }: TransactionHistoryItemProps) 
   const theme = useTheme();
   const explorerLink = getExplorerLink(transaction, currentNetworkConfig);
   const action = getTransactionAction(transaction);
-  const timestamp = getTransactionTimestamp(transaction);
+  const timestamp = Date.parse(transaction.timestamp);
 
   useEffect(() => {
     if (copyStatus) {
--- a/src/modules/history/TransactionRowItem.tsx
+++ b/src/modules/history/TransactionRowItem.tsx
@@ -11,12 +11,7 @@ import { GENERAL } from 'src/utils/events';
 import { useShallow } from 'zustand/shallow';
 
 import { ActionDetails, ActionTextMap } from './actions/ActionDetails';
-import {
-  getExplorerLink,
-  getTransactionAction,
-  getTransactionTimestamp,
-  unixTimestampToFormattedTime,
-} from './helpers';
+import { getExplorerLink, getTransactionAction, unixTimestampToFormattedTime } from './helpers';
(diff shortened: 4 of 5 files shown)


src/hooks/useTransactionHistory.tsx
```diff
@@ -108,89 +125,107 @@ export const useTransactionHistory = ({ isFilterActive }: { isFilterActive: bool
...
-        pool: selectedPool,
-      });
-      const cowSwapOrders = await fetchCowSwapsHistory(batchSize, skip * batchSize);
-      allTransactions.push(...currentBatch, ...cowSwapOrders);
-      currentBatchSize = currentBatch.length;
-      skip += batchSize;
-    }
+    const sdkTransactions = getSDKTransactions();
+
+    const allCowSwapOrders = await fetchCowSwapsHistory(PAGE_SIZE, 0);
```
foodaka: can we assign this 0 to a variable for clairty @AGMASO

foodaka:
@AGMASO is this clarified? "The reason I used the React hook directly is because I tried following the documentation to use the TypeScript action method, but it’s failing due import errors. Maybe the docs are not updated for the newer versions of the SDK?
"

src/hooks/useTransactionHistory.tsx
AGMASO: done

src/modules/history/types.ts
```diff
@@ -34,143 +36,81 @@ export type CowSwapSubset = {
...
-    collateralAssetPriceUSD: string;
-  } & PoolActionSubset;
-  SwapBorrowRate: {
-    reserve: ReserveSubset;
-    borrowRateModeFrom: string;
-    borrowRateModeTo: string;
-    stableBorrowRate: string;
-    variableBorrowRate: string;
-    assetPriceUSD: string;
-  } & PoolActionSubset;
```
mgrabina, line 77: arent these used for all the other history action items?

src/hooks/useTransactionHistory.tsx
```diff
@@ -108,89 +125,107 @@ export const useTransactionHistory = ({ isFilterActive }: { isFilterActive: bool
...
+    cursor: sdkCursor,
+  });
+
+  useEffect(() => {
+    setSdkCursor(null);
+    setSdkTransactions([]);
+    sdkTransactionIds.current.clear();
+    setIsFetchingAllSdkPages(true);
+    setHasLoadedInitialSdkPage(false);
+  }, [account, currentMarketData.addresses.LENDING_POOL, currentMarketData.chainId]);
```
mgrabina, line 156: do we need those useEffect deps?

src/hooks/useTransactionHistory.tsx
```diff
@@ -108,89 +125,107 @@ export const useTransactionHistory = ({ isFilterActive }: { isFilterActive: bool
...
+  useEffect(() => {
+    if (!sdkData?.items?.length) {
+      if (!sdkLoading && !sdkData?.pageInfo?.next) {
+        setIsFetchingAllSdkPages(false);
+        if (!hasLoadedInitialSdkPage) {
+          setHasLoadedInitialSdkPage(true);
+        }
+      }
+      return;
     }
```
mgrabina, line 168: we can probably simplify all pagination logic since we have 3 useEffects with 4 states

can prob use useInfiniteQuery

src/hooks/useTransactionHistory.tsx
```diff
@@ -377,6 +404,32 @@ export const useTransactionHistory = ({ isFilterActive }: { isFilterActive: bool
...
+      }
+      return page;
+    });
+
+    return {
+      ...data,
+      pages: pagesWithSdk,
+    };
+  }, [data, sdkTransactions]);
+
```
mgrabina, line 430: why we need this? can we just use a predefined hook like inifiniteQuery or just push new received intems in the same state array?

src/modules/history/types.ts
```diff
@@ -189,26 +129,38 @@ export interface HistoryFilters {
...
+      return FilterOptions.WITHDRAW;
+    case 'UserRepayTransaction':
+      return FilterOptions.REPAY;
+    case 'UserUsageAsCollateralTransaction':
     case 'CowCollateralSwap':
-      return 5;
-    case 'LiquidationCall':
-      return 6;
+      return FilterOptions.COLLATERALCHANGE;
+    case 'UserLiquidationCallTransaction':
```
mgrabina, line 142: Nice way more clear

src/modules/history/types.ts
```diff
@@ -189,26 +129,38 @@ export interface HistoryFilters {
 
 export const actionFilterMap = (action: string): number => {
   switch (action) {
-    case 'Deposit': // v2
```
mgrabina, line 192: are v2 removed intentionally?

src/hooks/useTransactionHistory.tsx
```diff
@@ -29,6 +34,15 @@ import { useShallow } from 'zustand/shallow';
 
 import { useAppDataContext } from './app-data-provider/useAppDataProvider';
 
+const sortTransactionsByTimestampDesc = (
+  a: TransactionHistoryItemUnion,
+  b: TransactionHistoryItemUnion
+) => {
+  const aTime = '__typename' in a ? new Date(a.timestamp).getTime() : a.timestamp * 1000;
+  const bTime = '__typename' in b ? new Date(b.timestamp).getTime() : b.timestamp * 1000;
+  return bTime - aTime;
```
mgrabina: so we are returning different timestamp formats from the Sdk compared to what cow or something else? i think best to adapt what we get from CoW or everywhere else (where we get those), so this functino is extremely clean or unnecessary

src/modules/history/types.ts
AGMASO: Yes, we already discussed that, the SDK isn’t pulling information from V2 transactions and because deprecating V2 on the interface.

src/modules/history/types.ts
AGMASO: The SDK already exports the required types for it to work. Those types don't match the old ones here.
For that reason, we now use the types provided by @aave/graphql. I refactored the `TransactionHistoryItemUnion` to align with the SDK types.

src/hooks/useTransactionHistory.tsx
AGMASO: done

src/hooks/useTransactionHistory.tsx
AGMASO: Yes, we need them: when `account`, `currentMarketData.addresses.LENDING_POOL`, or `currentMarketData.chainId` change, we need to reset the cursor, list, and flags.

src/hooks/useTransactionHistory.tsx
AGMASO: The problem was that I had to use the React hook from the SDK, `useUserTransactionHistory`. This comes with several limitations, since the React hook already encapsulates the request and doesn’t expose an async fetcher that can be passed to React Query. For this reason, this approach is the one that works. I used this hook because when I did this refactor, there was an error in the SDK documentation, which caused issues when trying to install the TypeScript version that does expose an async fetcher.

src/hooks/useTransactionHistory.tsx
AGMASO: Same explanation as before — since we’re using the React hook from the SDK, useUserTransactionHistory, we have some limitations. The useInfiniteQuery only fetches the CoW pages, while the transaction history from the SDK is obtained through the React hook. We need to merge both sources into the correct format. If we simply push the data into the array, the format won’t be correct.

PR #2693, opened 2025-10-02
feat: improve banner logic

foodaka:
## General Changes

- Improves logic for handling which banner is shown(before main net would briefly show and then the intended banner)
- Adds banner for arb limit orders

## Developer Notes

---

## Reviewer Checklist


src/layouts/MainLayout.tsx
```diff
@@ -120,18 +159,31 @@ const getCampaignConfigs = (
...
+  );
+  const { openSwitch } = useModalContext();
 
   const openMarket = (market: CustomMarket) => {
     setCurrentMarket(market);
     router.push(`/markets/?marketName=${market}`);
   };
 
-  const campaignConfigs = getCampaignConfigs(openMarket);
+  const campaignConfigs = getCampaignConfigs(openSwitch, openMarket);
```
mgrabina: would move getCampaignConfigs to a hook useBannerCampaigns so we can get openSwitch (or others in the future) right in there and just pass the chain id

src/layouts/MainLayout.tsx
```diff
@@ -3,21 +3,60 @@ import { Box } from '@mui/material';
...
+
+interface CampaignConfig {
+  notifyText: string;
+  buttonText: string;
+  buttonAction: ButtonAction;
+  bannerVersion: string;
+  icon: string;
+}
+
+type CampaignChainId = ChainId.base | ChainId.mainnet | ChainId.arbitrum_one;
```
mgrabina: why not just use generic ChainId if down here we use **Partial** of chains ids

src/layouts/MainLayout.tsx
```diff
@@ -3,21 +3,60 @@ import { Box } from '@mui/material';
...
 import { AppHeader } from './AppHeader';
-import TopBarNotify from './TopBarNotify';
+import TopBarNotify, { ButtonAction } from './TopBarNotify';
+
+interface CampaignConfig {
+  notifyText: string;
+  buttonText: string;
+  buttonAction: ButtonAction;
+  bannerVersion: string;
+  icon: string;
```
mgrabina: does it make sense to make the icon optional for future solo-text campaings

src/layouts/MainLayout.tsx
```diff
@@ -3,21 +3,60 @@ import { Box } from '@mui/material';
...
+    const selectedMarket = localStorage.getItem('selectedMarket');
+    if (selectedMarket && marketsData[selectedMarket as CustomMarket]) {
+      return marketsData[selectedMarket as CustomMarket].chainId;
+    }
+
+    // Priority 3: URL params marketName
+    const urlMarket = getQueryParameter('marketName');
+    if (urlMarket && marketsData[urlMarket as CustomMarket]) {
+      return marketsData[urlMarket as CustomMarket].chainId;
+    }
```
mgrabina: shouldn't this be prioritized e.g. someone opens a shared link but been using other network

src/layouts/MainLayout.tsx
```diff
@@ -73,16 +112,16 @@ const getCampaignConfigs = (
...
-  //   buttonText: 'Swap Now',
-  //   buttonAction: {
-  //     type: 'function' as const,
-  //     value: () => openSwitch('', ChainId.arbitrum_one),
-  //   },
-  //   bannerVersion: 'arbitrum-swap-v1',
-  //   icon: '/icons/networks/arbitrum.svg',
-  // },
+  [ChainId.arbitrum_one]: {
+    notifyText: 'Limit orders are now live on Arbitrum',
```
mgrabina: not sure how translations work but can we ensure banner messages will be translated this way? otherwise we can set notifyText type as react node and just use <Trans>

src/layouts/MainLayout.tsx
```diff
@@ -3,21 +3,60 @@ import { Box } from '@mui/material';
...
+  return ChainId.mainnet;
+};
 
 const getCampaignConfigs = (
-  // openSwitch: (underlyingAsset: string) => void,
+  openSwitch: (underlyingAsset: string, chainId: ChainId) => void,
   openMarket: (market: CustomMarket) => void
-) => ({
+): CampaignConfigs => ({
   [ChainId.base]: {
```
mgrabina: nit/v2: we can leverage vercel configs or amplitude feature flags to automatically change campaings without commiting new code

src/layouts/TopBarNotify.tsx
```diff
@@ -50,40 +51,57 @@ export default function TopBarNotify({ campaigns }: TopBarNotifyProps) {
...
+      const timer = setTimeout(() => {
+        setSlideIn(true);
+      }, 100);
+
+      return () => clearTimeout(timer);
+    } else {
+      setSlideIn(false);
     }
-  }, [currentCampaign, currentChainId]);
+  }, [showWarning]);
```
mgrabina, line 90: cant this be replaced by a showWarning fn that has the timeout inside?

src/layouts/MainLayout.tsx
mgrabina: i see 
`                <Trans>{currentCampaign.notifyText}</Trans>

`
just checking if it's enough

src/layouts/TopBarNotify.tsx
mgrabina, line 1: ui ideas 

Image

src/layouts/MainLayout.tsx
foodaka: yeah we dont really use translations much but added b2f5f50b2cdbae943041e5543e05e20905b5559f

src/layouts/MainLayout.tsx
foodaka: yes 33af44505321a41a5556963e627307309af354ad

src/layouts/MainLayout.tsx
foodaka: yeah good point 33af44505321a41a5556963e627307309af354ad

src/layouts/MainLayout.tsx
foodaka: done in 33af44505321a41a5556963e627307309af354ad

PR #2698, opened 2025-10-03
Refactor/reserve overview page sdk

AGMASO:
## General Changes

* Refactored `reserve-overview` to use the `aave-sdk`.
* We’re using the SDK across this page **except** for the Action modals. Refactoring those would require deeper changes to `ModalWrapper`, which would impact other modals not present on this page.

## Developer Notes

* This version isn’t 100% complete. We’re still missing a few fields that aren’t exposed by the SDK:

  1. `interestRateStrategyAddress` (ALREADY ADDED)
  2. Correct values for `isolationModeConfig.debtCeiling.amount` and `isolationModeConfig.totalBorrows.usd` (ALREADY ADDED)
  3. `reserve.unbacked` and `reserve.unbackedUSD` are not available in the SDK (SOLVED)

---

## Reviewer Checklist


pages/reserve-overview.page.tsx
```diff
@@ -120,11 +127,14 @@ export default function ReserveOverview() {
               width: { xs: '100%', lg: '416px' },
             }}
           >
-            <ReserveActions reserve={reserve} />
+            {/* Wrapped in AssetCapsProvider to provide the data using legacy method to avoid braking actions */}
```
foodaka, line 130: which breaking changes are there?

src/hooks/useAssetCapsSDK.tsx
```diff
@@ -0,0 +1,184 @@
...
+export const getBorrowCapData = (asset: ReserveWithId) => {
+  const totalDebt = valueToBigNumber(asset?.borrowInfo?.total.amount.value ?? '0');
+  const cap = valueToBigNumber(asset?.borrowInfo?.borrowCap.amount.value ?? '0');
+
+  const rawUsage = cap.isZero() ? 0 : totalDebt.dividedBy(cap).multipliedBy(100).toNumber();
+
+  const borrowCapReached = asset?.borrowInfo?.borrowCapReached || rawUsage >= 99.99;
+
+  return {
+    borrowCapUsage: Number.isFinite(rawUsage) ? rawUsage : 0,
```
foodaka, line 159: when does it go into the 0 block?

src/modules/reserve-overview/BorrowInfo.tsx
```diff
@@ -174,27 +183,32 @@ export const BorrowInfo = ({
           }
         >
           <IncentivesCard
-            value={reserve.variableBorrowAPY}
-            incentives={reserve.vIncentivesData || []}
-            address={reserve.variableDebtTokenAddress}
-            symbol={reserve.symbol}
+            value={
+              Number(reserve.borrowInfo?.total.amount.value) > 0
```
foodaka: i see we are checking the value here and then below an apy, is that correct?

AGMASO: you're right, it doesn't make much sense. I was thinking of it as a guard for cases where an asset has zero borrows, but it’s unnecessary. It's better to always show the APY if the SDK gives us a correctly formatted number.
I’ve just pushed a change to always display the borrow and supply APY when the value is valid. If we get NaN/undefined/Infinity, then we show NoData from the IncentiveCard.

src/hooks/useAssetCapsSDK.tsx
mgrabina: if the cap is zero and the supplied != 0 (e.g. after a gov freeze), shouldn't usage be 100%?

mgrabina (approved):

PR #2720, opened 2025-10-16
feat: added point based incentives for ink

AGMASO:
## General Changes

* Added new hook to handle point-based campaigns in Merkl
* Added a calculator so users can estimate the APY and USD value their points would generate.
* Calculated as well the FDV.
* Added campaign for the Tydro collaboration on Ink 

## Developer Notes

- Please check if these are the intended values for the points calculation:
I’ve calculated the following:

` const pointsPerThousandUsd = tvl > 0 ? (dailyPoints / tvl) * 1000 : 0;`

Screenshot 2025-10-29 at 16 36 49
Screenshot 2025-10-29 at 16 36 59

---

## Reviewer Checklist

### DIFF
--- a/src/components/infoTooltips/PointsBasedCampaignTooltip.tsx
+++ b/src/components/infoTooltips/PointsBasedCampaignTooltip.tsx
@@ -152,16 +152,23 @@ export const PointsBasedCampaignTooltip = ({
             backgroundColor: theme.palette.background.paper,
           }}
         >
-          <Typography variant={typographyVariant} color={theme.palette.text.secondary}>
-            Estimate APY and points value
-          </Typography>
-
+          {!hasInkPriceInput ? (
+            <Typography variant={typographyVariant} color={theme.palette.text.secondary}>
+              Calculate APY from the assumed INK price.
+              <br />
+              (INK price not public yet)
+            </Typography>
+          ) : (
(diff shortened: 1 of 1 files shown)


src/hooks/useMerklPointsIncentives.ts
```diff
@@ -0,0 +1,160 @@
...
+      }
+
+      const protocolIncentivesAPR = protocolIncentives.reduce((sum, inc) => {
+        return sum + (inc.incentiveAPR === 'Infinity' ? 0 : +inc.incentiveAPR);
+      }, 0);
+
+      const isBorrow = protocolAction === ProtocolAction.borrow;
+      const totalAPY = isBorrow
+        ? protocolAPY - protocolIncentivesAPR - merklIncentivesAPY
+        : protocolAPY + protocolIncentivesAPR + merklIncentivesAPY;
```
mgrabina: Are we sure we aren't mixing apr and apy which are two different things? in that case i recall a util convertAprToApy that mark used for apy history i think

AGMASO: You were right, we were mixing it up, but it was only affecting one token USDS since it’s the only one with `protocolIncentives` at the moment. I’ve fixed that.

PR #2725, opened 2025-10-21
fix: cooldown period updated

AGMASO:
## General Changes

- Fixed cooldown period for staking AAVE

## Developer Notes

---

## Reviewer Checklist

### DIFF
--- a/src/components/Warnings/CooldownWarning.tsx
+++ b/src/components/Warnings/CooldownWarning.tsx
@@ -1,3 +1,4 @@
+import { Stake } from '@aave/contract-helpers';
 import { Trans } from '@lingui/macro';
 import { Typography } from '@mui/material';
 import { useRootStore } from 'src/store/root';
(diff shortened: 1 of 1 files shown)


src/components/Warnings/CooldownWarning.tsx
```diff
@@ -7,7 +7,7 @@ import { Link } from '../primitives/Link';
 import { Warning } from '../primitives/Warning';
 import { SecondsToString } from '../SecondsToString';
 
-const TWENTY_DAYS = 20 * 24 * 60 * 60;
+const TWENTY_DAYS = 7 * 24 * 60 * 60; // 7 days
```
mgrabina: wanna change const name?

src/components/transactions/Stake/StakeModalContent.tsx
```diff
@@ -42,6 +42,7 @@ export const StakeModalContent = ({ stakeAssetName, icon }: StakeProps) => {
   const { data: stakeGeneralResult } = useGeneralStakeUiData(currentMarketData, stakeAssetName);
 
   const stakeData = stakeGeneralResult?.[0];
+  const stakeCooldownSeconds = stakeData?.stakeCooldownSeconds || 0;
```
mgrabina, line 45: we were just showing default right?

src/components/Warnings/CooldownWarning.tsx
AGMASO: ups

src/components/transactions/Stake/StakeModalContent.tsx
AGMASO: We were displaying a hardcoded value equal to 20 days. I updated to show the cooldown value coming from the SC, and giving a hardcoded fallback equal to 7 days

AGMASO: Actually, I’m thinking that the proposal only changed the cooldown period for AAVE, so I need to adjust the fallback to make sure it doesn’t affect other scenarios.

foodaka: can we read from the contract directly?

mgrabina (approved):

PR #2760, opened 2025-11-13
[BUG] fix: refactored sendTx to solve edgecase when using smart accounts erc20 paymaster txs

AGMASO:
## General Changes

* Fixed edge case when sending Txs using smart accounts erc20 paymaster(Metamask)
* Fixes: AAVEV3-7
### BUG DESCRIPTION

The incompatibility only arises when performing an action that requires `sendTx` using an EOA with an associated smart account in MetaMask **and** paying the fees using an ERC-20 Paymaster. In other words, the native token is *not* used to pay for the transaction.

In this scenario, the transaction **succeeds on-chain**, but the UI fails, causing confusion for the user. The balances are correctly updated, and if the user checks the transaction hash in MetaMask’s activity tab, they will see that the transaction was indeed executed successfully.

### CAUSE

The root cause is that ethers, in its `sendTransaction` function, does not account for the different flow that occurs when executing a smart account transaction using an ERC-20 Paymaster.

The flow works as follows:
The user submits the transaction using ethers’ `sendTransaction`. This transaction is intercepted by the smart account, which creates a UserOp with the corresponding values and sends it to the bundler. The bundler is the component that actually sends the transaction to the blockchain and produces a valid transaction hash.

Ethers internally fails to recognize this alternative flow—either due to unhandled logic or internal timeouts—which leads it to throw an error **even though the transaction was correctly sent to the bundler and the blockchain**.

### SOLUTION

We created a custom path to analyze the error emitted by ethers. If the error contains a valid transaction hash, it means that a smart account transaction with an ERC-20 Paymaster was executed. In that case, we extract the hash, fetch the corresponding receipt, and return it to the UI. This resolves the edge case.

If no valid hash exists in the error, we simply return the original error as usual.

## Reviewer Checklist

### DIFF
--- a/src/libs/web3-data-provider/Web3Provider.tsx
+++ b/src/libs/web3-data-provider/Web3Provider.tsx
@@ -96,11 +96,51 @@ export const Web3ContextProvider: React.FC<{ children: ReactElement }> = ({ chil
     if (provider) {
       const { from, ...data } = txData;
       const signer = provider.getSigner(from);
-      const txResponse: TransactionResponse = await signer.sendTransaction({
-        ...data,
-        value: data.value ? BigNumber.from(data.value) : undefined,
-      });
-      return txResponse;
+
+      try {
+        const txResponse: TransactionResponse = await signer.sendTransaction({
+          ...data,
+          value: data.value ? BigNumber.from(data.value) : undefined,
+        });
+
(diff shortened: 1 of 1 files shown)


mgrabina (approved):
good catch!!!

PR #2782, opened 2025-11-26
feat: new guard for borrowing power on debt swap

mgrabina:
adding new guard for the edge case when the borrowing power is not enough for the selected swap

image


src/components/transactions/Swap/errors/shared/InsufficientBorrowPowerBlockingGuard.tsx
```diff
@@ -0,0 +1,98 @@
...
+        };
+        setState({
+          error: blockingError,
+          actionsBlocked: {
+            [ActionsBlockedReason.INSUFFICIENT_BORROW_POWER]: true,
+          },
+        });
+      }
+    } else {
+      console.log('removing error');
```
foodaka: debug

foodaka (approved):

PR #2800, opened 2025-12-05
feat: activate swaps on linea and prepare plasma

mgrabina:


src/components/transactions/Swap/constants/cow.constants.ts
```diff
@@ -113,10 +113,12 @@ export const COW_UNSUPPORTED_ASSETS: Partial<
       '0xd190eF37dB51Bb955A680fF1A85763CC72d083D4'.toLowerCase(), // aGHO not supported
     ],
 
+    [SupportedChainId.LINEA]: [
+      '0xacA92E438df0B2401fF60dA7E4337B687a2435DA'.toLowerCase(), // mUSD, to test no provider supported, remove before prod
```
foodaka: should we remove then?

src/ui-config/marketsConfig.tsx
```diff
@@ -458,6 +465,13 @@ export const marketsData: {
       UI_INCENTIVE_DATA_PROVIDER: AaveV3Plasma.UI_INCENTIVE_DATA_PROVIDER,
       COLLECTOR: AaveV3Plasma.COLLECTOR,
     },
+    enabledFeatures: {
+      // liquiditySwap: true,
+      // withdrawAndSwitch: true,
+      // collateralRepay: true,
+      // debtSwitch: true,
+      // switch: true,
```
foodaka, line 473: is this supposed to be here for sonic?

foodaka:
verified addresses look good as well

src/ui-config/marketsConfig.tsx
mgrabina: This is plasma, prepared to turn on shortly

src/components/transactions/Swap/constants/cow.constants.ts
mgrabina: oh right, good catch thanks

PR #2807, opened 2025-12-10
fix: post hook gas limit calc for cow adapters swaps

mgrabina:
We found an issue for users that have multiple collaterals (usually multiple dust) where the calculations in the contracts for HF to allow the flashloan required excessive gas compared to most of the users. This fix is a hack to cover those cases, while CoW develops a more sustainable solution to estimating real gas needs via tx simulation.

Summary: we do BASE_GAS + 100k * collaterals_amount 

related https://github.com/cowprotocol/cow-sdk/pull/736


src/components/transactions/Swap/helpers/cow/adapters.helpers.ts
```diff
@@ -125,6 +125,18 @@ export const calculateInstanceAddress = async ({
...
 
+export const getHooksGasLimit = (
+  collateralsAmount: number
+): {
+  preHookGasLimit: bigint;
+  postHookGasLimit: bigint;
+} => {
+  return {
+    preHookGasLimit: BigInt(300000),
+    postHookGasLimit: BigInt(Math.min(600000 + collateralsAmount * 100000, 1500000)),
```
anxolin, line 136: I think you should add max, no?

mgrabina: Nope, I want to add max amount so we don't end up sending insane amounts.

mgrabina: So I need the min ( calculated, MAX )

PR #2824, opened 2025-12-30
feat: enable geo

foodaka:
## General Changes

- Fixes XYZ bug
- Adds XYZ feature
- …

## Developer Notes

---

## Reviewer Checklist

### DIFF
--- a/src/store/analyticsSlice.ts
+++ b/src/store/analyticsSlice.ts
@@ -95,7 +95,7 @@ export const createAnalyticsSlice: StateCreator<
             autocapture: true, // disable if we don't want to capture every click and page view on the site
             trackingOptions: {
               ipAddress: true,
-              language: false,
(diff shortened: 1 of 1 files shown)


AGMASO (approved):

src/store/analyticsSlice.ts
```diff
@@ -151,7 +151,7 @@ export const createAnalyticsSlice: StateCreator<
           init(AMPLITUDE_API_KEY, {
             autocapture: false,
             trackingOptions: {
-              ipAddress: false,
+              ipAddress: true,
```
mgrabina: maybe worth it to double check with legal if we can track IPs while opted out/not opted in yet

foodaka: yeah we should not, ive removed

PR #2830, opened 2026-01-06
fix: dust protection calculation optimization

mgrabina:
The dust protection (which was overestimated) was affecting the quote received only in one side, so for big orders where our overestimated costs are not significant and where slippage were low, the market orders were struggling more to get filled.

### DIFF
--- a/src/components/transactions/Swap/constants/cow.constants.ts
+++ b/src/components/transactions/Swap/constants/cow.constants.ts
@@ -151,7 +151,6 @@ export const COW_PARTNER_FEE = (tokenFromSymbol: string, tokenToSymbol: string) 
 });
 
 export const FLASH_LOAN_FEE_BPS = 5;
-export const VALID_TO_HALF_HOUR = Math.floor(Date.now() / 1000) + 60 * 10; // 10 minutes
 
 export const COW_APP_DATA = (
   tokenFromSymbol: string,


src/components/transactions/Swap/constants/cow.constants.ts
```diff
@@ -151,7 +151,7 @@ export const COW_PARTNER_FEE = (tokenFromSymbol: string, tokenToSymbol: string)
 });
 
 export const FLASH_LOAN_FEE_BPS = 5;
-export const VALID_TO_HALF_HOUR = Math.floor(Date.now() / 1000) + 60 * 30; // 30 minutes
```
mgrabina, line 154: No need for 30 min. Prefer lower amount to account for less dust.

src/components/transactions/Swap/constants/cow.constants.ts
```diff
@@ -60,7 +60,7 @@ export const ADAPTER_FACTORY: Record<SupportedChainId, string> = {
   [SupportedChainId.SEPOLIA]: '',
 };
 
-export const DUST_PROTECTION_MULTIPLIER = 1.001;
```
mgrabina, line 63: 👉 Conservative market
Max APR ≈ 80%
Max per hour: 0.91 bps

👉 Aggressive / volatile market
Max APR ≈ 300%
Max per hour:≈3.4 bps

👉 Extreme custom strategy (rare but allowed)
Max APR ≈ 1000%
Max per hour:≈11.4 bps

We have 10 min max now.

AGMASO (approved):

src/components/transactions/Swap/constants/cow.constants.ts
```diff
@@ -151,7 +151,7 @@ export const COW_PARTNER_FEE = (tokenFromSymbol: string, tokenToSymbol: string)
 });
 
 export const FLASH_LOAN_FEE_BPS = 5;
-export const VALID_TO_HALF_HOUR = Math.floor(Date.now() / 1000) + 60 * 30; // 30 minutes
+export const VALID_TO_HALF_HOUR = Math.floor(Date.now() / 1000) + 60 * 10; // 10 minutes
```
foodaka:
```suggestion
export const VALID_TO_TEN_MINUTES = Math.floor(Date.now() / 1000) + 60 * 10; // 10 minutes
```

mgrabina: good catch, realised not being used so removing

JoaquinBattilana (approved):

PR #2850, opened 2026-01-29
feat: gho on mantle

DhairyaSethi:
mantle network not supported on subgraph (https://thegraph.com/docs/en/supported-networks/) so txn page is broken
gho as feeToken is not supported by ccip on mantle yet, fee token through eth works fine


src/ui-config/marketsConfig.tsx
DhairyaSethi, line 1: market has no reserves yet but had to do this to test mantle to * bridge flows, perhaps there's a better workaround for unlisted markets?

PR #2868, opened 2026-02-10
fix: depentabot h3

mgrabina:
### Summary
Adds a Yarn resolution to force the transitive dependency `h3` to `^1.15.5`, fixing a critical HTTP Request Smuggling issue ([Dependabot alert](https://github.com/aave/interface/security/dependabot/106)).

### Background
- **Vulnerability:** In h3 ≤ 1.15.4, `Transfer-Encoding` is checked in a case-sensitive way. Requests with `Transfer-Encoding: ChuNked` (or other casing) are treated as having no body, which can lead to TE.TE desync and request smuggling when running behind L4 proxies or non-header-normalizing infrastructure.
- **Exposure:** We don’t depend on `h3` directly. It is pulled in transitively:
  - **wagmi** → **@wagmi/connectors** → **@walletconnect/ethereum-provider** → **@walletconnect/keyvaluestorage** → **unstorage** → **h3**
- Our lockfile had **h3@1.13.0** (vulnerable). The fix is in **h3@1.15.5+**.

### Approach
- **Chosen fix:** A single **resolutions** entry in `package.json` to force the whole dependency tree to use `h3@^1.15.5`.
- **Why resolutions:** We can’t “upgrade h3” as a direct dependency because it’s transitive. Resolutions tell Yarn to satisfy every `h3` request with the patched version, without upgrading wagmi/WalletConnect/unstorage and without risking breaking changes from a larger unstorage upgrade.
- **Risk:** Low — 1.15.5 is a patch release; no intentional API changes. Unstorage’s latest already uses `h3@^1.15.5`; we’re aligning the rest of the tree with that.

### Changes
- **package.json:** Added:
  ```json
  "resolutions": {
    "h3": "^1.15.5"
  }
  ```
- **yarn.lock:** Regenerated so all `h3` usages resolve to 1.15.5+.

### Follow-up
- When the WalletConnect/unstorage stack eventually depends on `h3@^1.15.5` and the lockfile naturally picks it up, this resolution can be removed.
- Resolves the Dependabot alert for `h3`.

Closes AAVE-3654

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,9 @@
   "version": "3.0.1",
   "private": true,
   "license": "BSD-3-Clause",
+  "resolutions": {
+    "h3": "^1.15.5"
+  },
   "scripts": {
     "dev": "next dev",
     "dev:special": "NODE_OPTIONS='--inspect --max_old_space_size=8000' next dev",
--- a/yarn.lock
+++ b/yarn.lock
@@ -6267,6 +6267,13 @@ cross-spawn@^7.0.0, cross-spawn@^7.0.2, cross-spawn@^7.0.3:
   dependencies:
     uncrypto "^0.1.3"
 
+crossws@^0.3.5:
+  version "0.3.5"
+  resolved "https://registry.yarnpkg.com/crossws/-/crossws-0.3.5.tgz#daad331d44148ea6500098bc858869f3a5ab81a6"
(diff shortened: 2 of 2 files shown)


mgrabina:
@codex

forhau (approved):

AGMASO (approved):

PR #2882, opened 2026-02-26
feat: added aptos rpc-proxy logic branch

mpsc0x:
## General Changes

- added aptos mainnet and testnet to proxy
- modified request type as aptos uses rest instead of jsonrpc

## Developer Notes

---

## Reviewer Checklist


pages/api/rpc-proxy.ts
```diff
@@ -88,14 +103,44 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
...
+    if (APTOS_CHAIN_IDS.has(chainIdNumber)) {
+      // Aptos uses REST API, not JSON-RPC
+      // Expected body: { chainId, method: 'GET' | 'POST', path: string, body?: object }
+      const { method, path, body } = req.body;
+
+      if (!path) {
+        return res.status(400).json({ error: 'Missing "path" for Aptos request' });
+      }
+
+      const aptosUrl = `${rpcUrl}/v1${path}`;
```
absis, line 123: No path sanitization. For example can be used like this:

```
{ "chainId": 90001, "method": "GET", "path": "/../../v2/some-other-key" }
```

Result: 
https://aptos-mainnet.g.alchemy.com/v2/<your-api-key>/v1/../../v2/some-other-key

Which after URL resolution could hit a different Alchemy endpoint using your API key

Posible solution:

```
if (!path.startsWith('/') || path.includes('..')) {
  return res.status(400).json({ error: 'Invalid path for Aptos request' });
}
```

absis (changes requested):

pages/api/rpc-proxy.ts
mpsc0x: @absis Fixed, pls check

absis (approved):

PR #2971, opened 2026-05-08
chore: set CoW flashloan fee to zero

mgrabina:
## Summary
Sets `FLASH_LOAN_FEE_BPS` to `0` in `src/components/transactions/Swap/constants/cow.constants.ts`. Paraswap's fee constant is unchanged.

The constant feeds CoW-based Collateral Swap, Debt Swap, and Repay-with-Collateral flows via `flashLoanSdk.calculateFlashLoanAmounts`. UI cost rows in `CowCostsDetails` already hide when the fee is zero.

### DIFF
--- a/src/components/transactions/Swap/constants/cow.constants.ts
+++ b/src/components/transactions/Swap/constants/cow.constants.ts
@@ -182,7 +182,7 @@ export const COW_PARTNER_FEE = (
   };
 };
 
-export const FLASH_LOAN_FEE_BPS = 5;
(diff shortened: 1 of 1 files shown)


mgrabina:
Switching to an on-chain ACLManager check approach (per-flow adapter, default to non-zero on pending). Will reopen with the new shape.

mgrabina:
Closing in favor of on-chain ACLManager check — see follow-up.

PR #2976, opened 2026-05-12
fix(swap): route EIP-7702 wallets to ParaSwap

mgrabina:
## Summary

Users on EIP-7702 delegated EOAs (Robinhood Wallet → Alchemy Modular Account v2, etc.) hit `InvalidEip1271Signature` from the CoW orderbook when running Aave's collateral swap / debt swap / repay-with-collateral / withdraw-and-swap flows.

Root cause: `@cowprotocol/sdk-flash-loans` hardcodes `SigningScheme.EIP1271` and a `customEIP1271Signature` that assumes `signer.signTypedData()` returns a plain ECDSA signature recoverable to the EOA. 7702 smart-account delegates intercept `signTypedData` at the wallet layer and return wrapped/module-encoded sigs, so the adapter's `isValidSignature` recovers the wrong address and CoW rejects. The simple EIP-712 `sendOrder` path likely breaks the same way on these wallets.

This change detects 7702 delegation (code prefix `0xef0100`) in the swap user context and forces ParaSwap for those users — ParaSwap is on-chain and doesn't depend on off-chain signature verification, so it works for any wallet regardless of delegate behavior.

## Changes

- `src/helpers/provider.ts`: export `isEip7702Wallet` helper (wraps the existing `isEip7702EOA` predicate)
- `src/components/transactions/Swap/types/state.types.ts`: add `userIsEip7702Wallet` to `SwapState`
- `src/components/transactions/Swap/hooks/useUserContext.ts`: detect 7702 alongside `isSmartContractWallet` / `isSafeWallet`
- `src/components/transactions/Swap/helpers/shared/provider.helpers.ts`: `getSwitchProvider` skips CoW when `userIsEip7702Wallet`
- `src/components/transactions/Swap/hooks/useSwapQuote.ts`: thread the flag through

## Repro

Support ticket: wallet `0x50fcB386A45633cda069dB74c64f7E8c2bf90C56` on BSC, BNB→USDT collateral swap, returned order hash `0xb450a1d3b86bc9be05a7f47ae356a06788ec82ad45180c20a1bcb6bd122f7364`. Delegate at `0x69007702764179f14F51cdce752f4f775d74E139` is Alchemy Modular Account v2.

## Follow-ups

- Flagged the SDK-side incompatibility to CoW; long-term fix is ERC-6492 / 7702-aware validation in `@cowprotocol/sdk-flash-loans` so we can re-enable CoW for these wallets.
- Worth a sanity check whether the second branch of `isEip7702EOA` (`code === account.toLowerCase()`) was intentional — it doesn't match any standard 7702 delegation pattern. Not changed here.

## Test plan

- Manual: connect a 7702-delegated wallet (Robinhood Wallet, or MetaMask Smart Account on Sepolia) and verify the swap UI picks ParaSwap, not CoW, for collateral / debt / withdraw / repay-with-collateral flows.
- Regression: connect a plain EOA and a Safe wallet, verify provider selection is unchanged (CoW for EOA where supported, PRESIGN/Safe path for Safe).

### DIFF
--- a/src/components/transactions/Swap/actions/approval/useSwapTokenApproval.ts
+++ b/src/components/transactions/Swap/actions/approval/useSwapTokenApproval.ts
@@ -109,7 +109,11 @@ export const useSwapTokenApproval = ({
   const { sendTx, signTxData } = useWeb3Context();
   const [loadingPermitData, setLoadingPermitData] = useState(true);
 
-  const { isSmartContractWallet, isLoading: isLoadingWalletType } = useGetConnectedWalletType();
+  const {
+    isSmartContractWallet,
+    isEip7702Wallet,
+    isLoading: isLoadingWalletType,
+  } = useGetConnectedWalletType();
 
   const [
     user,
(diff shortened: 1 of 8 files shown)


mgrabina:
@codex

PR #2983, opened 2026-05-15
fix(web3): use useAccountEffect for walletType to fix tab-switch slowdown

mgrabina:
## Summary

#2974 destructured `status` and `connector` from `useAccount()` to drive `walletType` from wagmi's account state. That destructure is what slowed `app.aave.com` down.

Wagmi v2's `useAccount` uses `useSyncExternalStoreWithTracked` ([source](https://github.com/wevm/wagmi/blob/main/packages/react/src/hooks/useSyncExternalStoreWithTracked.ts)): each property you destructure is pushed into a tracked-key set, and the snapshot equality function deep-equals every tracked key on every wagmi store tick. Adding `connector` to that set forced a per-tick `deepEqual` over a multi-property object with an internal event emitter, re-rendering `Web3ContextProvider` on every wagmi update. Since `Web3Context.Provider` builds its `value` object fresh on each render, every context consumer re-rendered too, which is what tab switching paid for.

The replacement: `useAccountEffect`, which calls `watchAccount` inside a plain `useEffect` and does not subscribe the component to renders. It fires `onConnect` for both fresh connect and reconnect (page reload, redirect-back), so the original analytics fix from #2974 still holds.

## Verification

Local A/B with Playwright. 20 samples per variant, dev mode, wallet disconnected, top-nav click → URL commit → next paint:

| variant | commit | min | **p50** | p95 | mean |
|---|---|---|---|---|---|
| main before #2974 | 06a69eb77 | 18 | **24** | 100 | 31 |
| main with #2974 (today) | b0890e211 | 52 | **66** | 222 | 84 |
| this branch | — | 19 | **25** | 95 | 32 |

I bisected inside #2974: reverting just the `status, connector` destructure restores baseline; removing only the new `useEffect` does not. So the destructure is the cause, not the effect.

## Out of scope

Other consumers in the codebase that destructure a wide field set from `useAccount()` may have the same shape of overhead. Not changing them here, but worth a sweep.

### DIFF
--- a/src/libs/web3-data-provider/Web3Provider.tsx
+++ b/src/libs/web3-data-provider/Web3Provider.tsx
@@ -207,12 +207,8 @@ export const Web3ContextProvider: React.FC<{ children: ReactElement }> = ({ chil
     setAccount(account?.toLowerCase());
   }, [account, setAccount]);
 
-  // Drive walletType from wagmi's account events. useAccountEffect fires for both
-  // fresh connects and reconnects (page reload, redirect-back), so analytics keeps
-  // a real connector id instead of falling back to undefined after a refresh.
-  // It uses watchAccount under the hood and does not subscribe this component to
-  // account state, so destructuring extra fields here would force wagmi's tracked
-  // subscription to deep-equal the connector object on every store tick.
+  // useAccountEffect fires on connect AND reconnect (so analytics survives reloads)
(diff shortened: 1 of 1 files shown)


mgrabina:
@codex

grothem (approved):

sammdec (approved):

PR #2998, opened 2026-05-26
feat: add Narval wallet connector

mattschoch:
## General Changes

- Adds an optional Narval wallet connector to the connect modal to enable BitGo wallets, gated on a new `NEXT_PUBLIC_NARVAL_CLIENT_ID` env var. When the env var is unset (the default for the public app), the connector is not registered and behavior is unchanged.
- Preserves the existing Safe `allowedDomains` configuration and the "family connector sorted last" workaround in both modes (with and without Narval enabled).
- Plumbs `NEXT_PUBLIC_NARVAL_CLIENT_ID` through `.env.example`, `.env.development`, `.github/actions/build/action.yml`, and the prod/dev deploy workflows so a maintainer can set it via repo secrets when ready.

## Developer Notes

- **New dependency:** [`@narval-xyz/connect@^0.0.8`](https://www.npmjs.com/package/@narval-xyz/connect). Installed via `yarn add`; lockfile updated. Transitive additions: `ox@^0.9.12`, `zod@^4.3.6`.
- **No-op when unconfigured:** `narvalConnector` is only constructed when `process.env.NEXT_PUBLIC_NARVAL_CLIENT_ID` is truthy. Builds without the secret (e.g. forks, the staging build job, `build-fork.yml`) skip Narval entirely — no runtime change vs. \`main\`.
- **Maintainer action required to enable in production:** the value for \`NEXT_PUBLIC_NARVAL_CLIENT_ID\` needs to be added to the repo's GitHub Actions secrets (same name) before it will be embedded in deployed bundles. Until then, the workflows resolve it to an empty string and Narval stays disabled — safe default.
- The diff to \`wagmiConfig.ts\` deliberately keeps the existing \`prodCkConfig.connectors\` map/sort pipeline intact and only prepends \`narvalConnector\` to its result, so the Safe and Family invariants documented by the inline comments are preserved when Narval is enabled.

---

## Reviewer Checklist

- [ ] End-to-end tests are passing without any errors
- [ ] Code changes do not significantly increase the application bundle size
- [ ] If there are new 3rd-party packages, they do not introduce potential security threats
- [x] If there are new environment variables being added, they have been added to the \`.env.example\` file as well as the pertinant \`.github/actions/*\` files
- [ ] There are no CI changes, or they have been approved by the DevOps and Engineering team(s)


src/ui-config/wagmiConfig.ts
```diff
@@ -104,6 +105,15 @@ const connectors = prodCkConfig.connectors
...
   });
 
+const narvalConnector = process.env.NEXT_PUBLIC_NARVAL_CLIENT_ID
+  ? narval({
+      config: {
+        clientId: process.env.NEXT_PUBLIC_NARVAL_CLIENT_ID,
+      },
+    })
+  : undefined;
+const connectors = narvalConnector ? [narvalConnector, ...(baseConnectors ?? [])] : baseConnectors;
```
sammdec: Would it be possible to append to the connector list rather than prepend https://family.co/docs/connectkit/wallets

sammdec (changes requested):

src/ui-config/wagmiConfig.ts
mattschoch: The order still doesn't change on the UI, ConnectKit reorders it still.

Code here:
https://github.com/family/connectkit/blob/025b3f4b643e9bb8828d63fcbd068f182b64f60b/packages/connectkit/src/wallets/useWallets.tsx#L109-L119

ConnectKit makes the default connectors always at the bottom, and then sorts any connectors with `type: 'injected'` to the top.

I've changed our connector type to a custom `narval` type, but ConnectKit displays an error screen for that:

In `useWallets.tsx`, ConnectKit overwrites any installed field with:
```
  isInstalled:
    connector.type === 'mock' ||
    (connector.type === 'injected' && connector.id !== 'metaMask') ||
    isCoinbaseWalletConnector(connector.id),  // id === 'coinbaseWalletSDK'
```

  Then in ConnectWithInjector/index.tsx:
```
  const [status, setStatus] = useState(
    forceState ? forceState : !wallet?.isInstalled ? states.UNAVAILABLE : states.CONNECTING
  );
  ...
  const runConnect = async () => {
    if (wallet?.isInstalled && wallet?.connector) {
      connect({ connector: wallet?.connector });
    } else {
      setStatus(states.UNAVAILABLE);  // shows "Unsupported Browser"
    }
  };
```
  So by moving off type: 'injected', it treats isInstalled as false, and ConnectKit now thinks it's an "install the browser extension" wallet.

---
I see 2 options
1. We keep it as `injected` which ConnectKit will order higher on the list.
2. I open a PR to fix the ordering on ConnectKit, but that requires ConnectKit to accept it/publish & then we need to bump the dep here too.

wdyt?

mattschoch: Opened the fix PR on ConnectKit here: https://github.com/family/connectkit/pull/511

If that can approved, I can update this PR with the updated connectkit version.

sammdec (approved):

grothem:
thanks @mattschoch, i'm going to merge this to a staging branch so we can run some final tests before merging to main

PR #3010, opened 2026-06-05
feat: funkit checkout integration for Core-market supplies

yogurtandjam:
## What

Routes Supply for 4 allowlisted assets (USDC, USDT, WBTC, cbBTC) on the **Core mainnet market** through the [funkit](https://docs.fun.xyz) checkout modal, letting users supply from any EVM asset or fiat. All other assets/markets keep the native Aave supply flow.

### Integration architecture
- **`FunkitCheckout`** — `ssr:false` modal host mounted once in `_app` next to the app's other modal hosts (`SupplyModal` etc.). Owns the single `useFunkitCheckout` instance; per-asset configs are passed at call time via `beginCheckout(configOverride)`. Reuses the app's wagmi + react-query + ConnectKit (no funkit wallet stack). Honors the SDK's `onLoginFinished` resume contract by replaying it once ConnectKit connects.
- **`funSupplyBridge`** — module-level imperative bridge across the client-only island boundary (`@funkit/connect` is browser-only ESM; the app pre-renders/static-exports). The island registers `beginSupply` on mount; Supply buttons invoke it at click time. Clicks before the chunk loads fall back to the native modal instead of dropping.
- **`useSupplyButtonAction`** — single shared click handler for all 3 Supply list-item variants; gates on `isFunSupplyAsset(currentMarket, underlyingAsset)`.

### Data sourcing — no integrator-owned copies
- The only hardcoded data is **`FUN_SUPPLY_UNDERLYINGS`: a 4-address allowlist Set** — purely the product decision of which assets route through fun checkout.
- Receipt-token metadata (aToken address, real on-chain symbol `aEthUSDC`, decimals, hosted icon) comes from **app state**: `useAppDataContext().supplyReserves` (`@aave/react` SDK reserves whose `aToken`/`underlyingToken` are `Currency` objects) — the same source `AddTokenDropdown` renders on reserve-overview. Display fields (symbol, APY, collateral state) come from the clicked dashboard reserve; pool address from `currentMarketData`.
- Gated on `CustomMarket.proto_mainnet_v3` (not chainId): mainnet hosts three markets (Core/Prime/EtherFi) and e.g. USDC exists in all three with different aTokens and pools. Prime/EtherFi supplies of the same underlyings stay native. If SDK market data hasn't loaded at click time, the click falls back to the native modal.

### Add-to-wallet icon parity
The native flow's wallet icon is *generated*, not hosted: `Base64Token` wraps the underlying's local SVG in Aave's gradient `TokenRing` and base64-encodes it. Fun-routed supply rows mount the same hidden generator via the shared **`useFunSupplyATokenIcon`** hook (the `AddTokenDropdown` pattern) and pass the data URI through the click payload — ready at row mount, never delays `beginCheckout`. The SDK's hosted `aToken.imageUrl` is the fallback for a click that beats generation.

### Zero-balance behavior
fun-routed assets always render in "Assets to supply" and keep an enabled Supply button at 0 wallet balance (users can pay with any asset/fiat). Protocol-level blocks (inactive/frozen/paused/capped) still disable.

### Styling shim
`funkitPreflight.css` — a zero-specificity (`:where()`) slice of Tailwind preflight scoped under funkit's `[data-rk]` portal root. `@funkit/connect` implicitly relies on host Tailwind preflight (its production hosts are Tailwind apps) for form-control font inheritance and `img { max-width: 100% }`; this app is MUI/emotion. Without it the amount-input digits render in system font at weight 400 and host-supplied icons paint at intrinsic size.

### Compat
- `patches/@funkit+connect+9.18.0.patch` — React 18 `forwardRef` shim for funkit's React 19 ref-as-prop components (patch-package, wired via postinstall).
- `.yarnrc` `ignore-engines` — `@solana/addresses` (transitive) requires node ≥20.18 while the repo targets node 18; browser-only code, engines gate is install-only.

## Not in scope (flagged for follow-up)
- Reserve-overview page Supply button (`ReserveActions.tsx`) and SuppliedPositions "supply more" buttons still go native — product decision pending.
- Permit toggle for the Aave native-supply fast path: tracked internally at fun.xyz (ENG-4068).
- funkit SDK: lazy `addToWalletToken.iconSrc` resolver (consume-time semantics) would remove the begin-time icon coupling for all integrators.

## Test plan
- [ ] Core market: Supply USDC/USDT/WBTC/cbBTC from dashboard → funkit modal opens immediately, themed correctly (light+dark), amount input renders in Inter at correct weight
- [ ] Modal shows "You receive … aEthUSDC" (SDK-sourced ticker) with the ringed aToken icon
- [ ] Checkout complete: destination row shows 15px Aave icon; "Add aEthUSDC to your balance" registers the gradient-ringed icon in MetaMask with no symbol-mismatch warning
- [ ] Prime/EtherFi markets: Supply USDC → native Aave modal
- [ ] Non-allowlisted assets: native modal
- [ ] Empty wallet on Core: 4 fun assets visible with enabled Supply buttons
- [ ] Disconnected: Supply click opens ConnectKit


.yarnrc
sammdec, line 1: @yogurtandjam This can be deleted now as our nvrmrc requires node version 20.19

.prettierignore
```diff
@@ -26,3 +26,5 @@ src/locales/
 *.mermaid
 *.log
 *.lock
+*.patch
```
sammdec: @yogurtandjam Do we still need this patch?

.prettierignore
```diff
@@ -26,3 +26,5 @@ src/locales/
...
+.yarnrc
```
sammdec: Can we delete this?

.gitignore
```diff
@@ -45,7 +46,7 @@ package-lock.json
 .yarn-cache
 
 # TypeScript incremental build cache
-tsconfig.tsbuildinfo
+*.tsbuildinfo
```
sammdec: @yogurtandjam Why do we need this?

package.json
```diff
@@ -140,6 +146,7 @@
     "jest": "^29.7.0",
     "jest-environment-jsdom": "^29.7.0",
     "lint-staged": "^13.0.3",
+    "patch-package": "^8.0.1",
```
sammdec: Can we remove this as it seems like the patch has been removed

package.json
```diff
@@ -29,6 +30,7 @@
     "i18n:extract": "NODE_ENV=development lingui extract --clean --overwrite --locale en",
     "i18n:compile": "lingui compile",
     "i18n": "yarn i18n:extract && yarn i18n:compile",
+    "postinstall": "patch-package",
```
sammdec: Can we delete

sammdec (changes requested):
Some small initial changes requested

.gitignore
yogurtandjam: ah, local dev testing created some differently named tsconfig build files so i generalized the ignore statement and just forgot to change this back, mb 🙇

sammdec (approved):

PR #3015, opened 2026-06-15
feat: funkit checkout integration for Core-market supplies (#3010)

sammdec:
See #3010 for original PR

## What

Routes Supply for 4 allowlisted assets (USDC, USDT, WBTC, cbBTC) on the **Core mainnet market** through the [funkit](https://docs.fun.xyz) checkout modal, letting users supply from any EVM asset or fiat. All other assets/markets keep the native Aave supply flow.

### Integration architecture
- **`FunkitCheckout`** — `ssr:false` modal host mounted once in `_app` next to the app's other modal hosts (`SupplyModal` etc.). Owns the single `useFunkitCheckout` instance; per-asset configs are passed at call time via `beginCheckout(configOverride)`. Reuses the app's wagmi + react-query + ConnectKit (no funkit wallet stack). Honors the SDK's `onLoginFinished` resume contract by replaying it once ConnectKit connects.
- **`funSupplyBridge`** — module-level imperative bridge across the client-only island boundary (`@funkit/connect` is browser-only ESM; the app pre-renders/static-exports). The island registers `beginSupply` on mount; Supply buttons invoke it at click time. Clicks before the chunk loads fall back to the native modal instead of dropping.
- **`useSupplyButtonAction`** — single shared click handler for all 3 Supply list-item variants; gates on `isFunSupplyAsset(currentMarket, underlyingAsset)`.

### Data sourcing — no integrator-owned copies
- The only hardcoded data is **`FUN_SUPPLY_UNDERLYINGS`: a 4-address allowlist Set** — purely the product decision of which assets route through fun checkout.
- Receipt-token metadata (aToken address, real on-chain symbol `aEthUSDC`, decimals, hosted icon) comes from **app state**: `useAppDataContext().supplyReserves` (`@aave/react` SDK reserves whose `aToken`/`underlyingToken` are `Currency` objects) — the same source `AddTokenDropdown` renders on reserve-overview. Display fields (symbol, APY, collateral state) come from the clicked dashboard reserve; pool address from `currentMarketData`.
- Gated on `CustomMarket.proto_mainnet_v3` (not chainId): mainnet hosts three markets (Core/Prime/EtherFi) and e.g. USDC exists in all three with different aTokens and pools. Prime/EtherFi supplies of the same underlyings stay native. If SDK market data hasn't loaded at click time, the click falls back to the native modal.

### Add-to-wallet icon parity
The native flow's wallet icon is *generated*, not hosted: `Base64Token` wraps the underlying's local SVG in Aave's gradient `TokenRing` and base64-encodes it. Fun-routed supply rows mount the same hidden generator via the shared **`useFunSupplyATokenIcon`** hook (the `AddTokenDropdown` pattern) and pass the data URI through the click payload — ready at row mount, never delays `beginCheckout`. The SDK's hosted `aToken.imageUrl` is the fallback for a click that beats generation.

### Zero-balance behavior
fun-routed assets always render in "Assets to supply" and keep an enabled Supply button at 0 wallet balance (users can pay with any asset/fiat). Protocol-level blocks (inactive/frozen/paused/capped) still disable.

### Styling shim
`funkitPreflight.css` — a zero-specificity (`:where()`) slice of Tailwind preflight scoped under funkit's `[data-rk]` portal root. `@funkit/connect` implicitly relies on host Tailwind preflight (its production hosts are Tailwind apps) for form-control font inheritance and `img { max-width: 100% }`; this app is MUI/emotion. Without it the amount-input digits render in system font at weight 400 and host-supplied icons paint at intrinsic size.

### Compat
- `patches/@funkit+connect+9.18.0.patch` — React 18 `forwardRef` shim for funkit's React 19 ref-as-prop components (patch-package, wired via postinstall).
- `.yarnrc` `ignore-engines` — `@solana/addresses` (transitive) requires node ≥20.18 while the repo targets node 18; browser-only code, engines gate is install-only.

## Not in scope (flagged for follow-up)
- Reserve-overview page Supply button (`ReserveActions.tsx`) and SuppliedPositions "supply more" buttons still go native — product decision pending.
- Permit toggle for the Aave native-supply fast path: tracked internally at fun.xyz (ENG-4068).
- funkit SDK: lazy `addToWalletToken.iconSrc` resolver (consume-time semantics) would remove the begin-time icon coupling for all integrators.

## Test plan
- [ ] Core market: Supply USDC/USDT/WBTC/cbBTC from dashboard → funkit modal opens immediately, themed correctly (light+dark), amount input renders in Inter at correct weight
- [ ] Modal shows "You receive … aEthUSDC" (SDK-sourced ticker) with the ringed aToken icon
- [ ] Checkout complete: destination row shows 15px Aave icon; "Add aEthUSDC to your balance" registers the gradient-ringed icon in MetaMask with no symbol-mismatch warning
- [ ] Prime/EtherFi markets: Supply USDC → native Aave modal
- [ ] Non-allowlisted assets: native modal
- [ ] Empty wallet on Core: 4 fun assets visible with enabled Supply buttons
- [ ] Disconnected: Supply click opens ConnectKit


JoaquinBattilana (dismissed):

src/ui-config/funkit/funkitConfig.tsx
```diff
@@ -0,0 +1,42 @@
...
+ * `utils/constants/funkit.ts`) and the funkit playground's `aave` customer
+ * (`with-next/configs/customers.tsx`). The matching theme lives in
+ * `./aaveTheme` (ported from `with-next/themes/aave.ts`).
+ *
+ * The API key comes from `NEXT_PUBLIC_FUNKIT_API_KEY` (per the repo no-secrets
+ * rule); checkout is non-functional without it.
+ */
+export const funkitConfig: FunkitConfig = {
+  appName: 'Aave',
+  apiKey: process.env.NEXT_PUBLIC_FUNKIT_API_KEY || '',
```
mgrabina, line 19: I think we should prefer to use clientEnv/serverEnv for typed env vars

src/components/transactions/FunCheckout/funSupplyAssets.ts
```diff
@@ -0,0 +1,106 @@
...
+ */
+const FUN_SUPPLY_MARKET = CustomMarket.proto_mainnet_v3;
+
+/**
+ * The product allowlist — the underlyings (lowercased) whose Supply button
+ * routes through funkit's checkout modal instead of the native Aave supply
+ * modal. Everything else about these assets (symbols, decimals, addresses,
+ * icons) comes from live app state at click time.
+ */
+export const FUN_SUPPLY_UNDERLYINGS: ReadonlySet<string> = new Set([
```
mgrabina, line 21: should we move this to a LD flag?

src/components/transactions/FunCheckout/FunkitCheckout.tsx
```diff
@@ -0,0 +1,176 @@
...
+  }, [address]);
+
+  const colorMode = muiTheme.palette.mode;
+
+  useEffect(() => {
+    const persisted = (localStorage?.getItem('colorMode') as 'light' | 'dark') || colorMode;
+    toggleTheme(persisted);
+  }, [colorMode, toggleTheme]);
+
+  const beginSupply = async (reserve: FunSupplyReserve) => {
```
mgrabina, line 189: should we add analytics to track usage and/or conversion rates of funkit vs normal supply flow

mgrabina (dismissed):

src/components/transactions/FunCheckout/funSupplyAssets.ts
yogurtandjam: !! by this do you mean launch darkly? if y'all use a feature gating service i'd definitely recommend putting this behind one. i couldn't find any featuregating code or packages in the repo so i made this PR using an env flag https://github.com/aave/interface/pull/3031 but LD is much better

src/components/transactions/FunCheckout/FunkitCheckout.tsx
yogurtandjam: do you have analytics on your side? funkit tracks conversion rates per customer

src/ui-config/funkit/funkitConfig.tsx
yogurtandjam: yup, do we have an existing pattern for this in the codebase? or is this a new pattern that your team will set up

src/components/transactions/FunCheckout/funSupplyAssets.ts
sammdec: I don't think we have LD integrated into this repo yet @mgrabina

src/components/transactions/FunCheckout/FunkitCheckout.tsx
sammdec: @mgrabina do we already have normal conversion tracking? If so we could just grab the funkit conversion data and measure from there. Ideally we could run as an experiment in LD

src/ui-config/funkit/funkitConfig.tsx
sammdec: This would be a new pattern I think its ok for now to continue as is what do you think @mgrabina

src/components/transactions/FunCheckout/FunkitCheckout.tsx
sammdec: @mgrabina added events in this commit https://github.com/aave/interface/pull/3015/commits/81e614634429dbafa2d96cf2167fad9466c0540a

AGMASO (approved):

JoaquinBattilana (approved):

PR #3018, opened 2026-06-15
fix: inject NEXT_PUBLIC_FUNKIT_API_KEY in CI builds; guard FunkitCheckout against a missing key

yogurtandjam:
The funkit checkout host is mounted globally in _app, and FunkitProvider's validateFunkitConfig throws 'Invalid funkitConfig: Missing apiKey.' on an empty apiKey. The CI build pipeline never injected NEXT_PUBLIC_FUNKIT_API_KEY, so every cypress smoke build crashed the whole app on load (all markets) — the dashboard toggle and supply modal never rendered.

- Inject NEXT_PUBLIC_FUNKIT_API_KEY (from the new FUNKIT_API_KEY repo secret) through the build composite action and every workflow that builds the app (build-test-deploy, build-test-deploy-dev, build-fork, build_staging).
- Guard FunkitCheckout: render null when funkitConfig.apiKey is empty, so a missing/misconfigured key can never white-screen the app again — Supply falls back to the native modal via the existing beginFunSupply()->false path.
- Revert the temporary DebugCrashBoundary (#3017) used to pinpoint this.

Requires the FUNKIT_API_KEY secret to be added to the repo for funkit to be functional (publishable client-side key); without it the app degrades to native supply rather than crashing.


.github/workflows/build-fork.yml
```diff
@@ -22,6 +22,7 @@ jobs:
         uses: ./.github/actions/build
         with:
           NEXT_PUBLIC_ENV: 'prod'
+          NEXT_PUBLIC_FUNKIT_API_KEY: ${{ secrets.FUNKIT_API_KEY }}
```
yogurtandjam: i am not sure we need this tbh, just being comprehensive. i think we only *need* it for test-deploy jobs

PR #3024, opened 2026-06-16
fix: poll pool queries during funkit checkout for cross-chain updates

yogurtandjam:
## Summary
- After completing a "transfer crypto" (cross-chain bridge) in the FunKit modal, the dashboard wasn't updating because `onSuccess` fires before bridged funds are indexed on the destination chain
- Mirrors Polymarket's pattern: polls pool queries every 5 seconds while the FunKit checkout is open, stopping on close or success
- Adds `onClose` callback to clear the polling interval when the modal is dismissed

## Test plan
- [ ] Open the app, connect wallet, click Supply on a FunKit-eligible asset (USDC, USDT, WBTC, cbBTC on mainnet)
- [ ] Choose "Transfer crypto" in the FunKit modal and complete a bridge transfer
- [ ] Confirm the dashboard updates within ~5 seconds of funds arriving
- [ ] Close the modal without completing a tx — verify no dangling polling interval in DevTools
- [ ] Test same-chain payment flows (card, wallet) still work as before


yogurtandjam:
Superseded by #3025 (targets feat/funkit-integration instead of main)

PR #3063, opened 2026-07-03
feat: governance cache SDK, payloads panel, and gasless voting

JoaquinBattilana:
## Summary

Migrates governance/proposal pages off the TheGraph subgraph onto the governance-v3-cache GraphQL API, adds a payloads panel with executable targets and Seatbelt links, revives gasless voting behind a server-side Gelato relay, and adds an experimental redesigned proposal timeline.

### Governance cache SDK
- New typed fetching layer in `src/services/governance-cache-sdk/` (raw fetch + `gql` authoring + hand-written types) over the curated cache endpoints, plus the auto-generated/derived fields needed to fully replace the subgraph.
- React Query hooks consolidated into `src/hooks/governance/useGovernanceCache.ts`; query keys centralized in `queryKeysFactory.governanceCache*`.
- Old subgraph data path removed from governance.

### Payloads panel
- New `ProposalPayloads` panel: per-payload chain, id, state, access level, creator, controller, executable targets, and Seatbelt report link.
- Lifecycle panel heading renamed to "Timeline"; the Seatbelt link moved into the payloads panel.
- SDK now surfaces `PayloadAction[]` (target/signature/callData/value/withDelegateCall/accessLevel) per payload.

### Gasless voting
- New `pages/api/gelato/relay.ts` server route holds the Gelato sponsor key server-side (never bundled to the client) and relays via Gelato's sponsored-call REST endpoint.
- The route whitelists the target against known voting machines and requires the `submitVoteBySignature` selector, so sponsored gas can't be spent on arbitrary transactions.
- Client gasless path is a single step (sign EIP-712 -> relay -> poll public task status); the dead two-step approval scaffolding was removed.
- Gated by `NEXT_PUBLIC_ENABLE_GASLESS_VOTING` (public on/off) plus the server-side `GELATO_SPONSOR_KEY`; dormant unless both are set.

### Proposal timeline
- New `src/modules/governance/proposal/ProposalTimeline.tsx`: a single-spine timeline that weaves the three state machines (proposal core / voting machine / per-chain payloads) onto one line, driven entirely by existing cache data.
- Each step renders a date when done, a **live countdown** while it's the active wait (creation cooldown, voting window, execution/payload timelocks), or a **"Ready to X"** badge when a permissionless crank is due.
- Future steps show a forward-**projected approximate date** (`~`) chained from the known durations, floored at `now` so a step that's ready-but-awaiting-its-tx never reads as a past/overdue date.
- Every underlying transaction is a plain-language, collapsible substep with a block-explorer link slot; per-chain payload breakdown lives in the expand.
- Payload outcomes: green when all executed, solid yellow when settled with some expired/cancelled (e.g. `2 / 3 executed · 1 expired`), hollow while still waiting.
- **Replaces** the previous `ProposalLifecycleCache` (now deleted) as the sole proposal timeline; the `Forum discussion` button moved from the timeline to the `Voting results` panel.
- **Wired to real data** (depends on the governance-v3-cache changes in aave/governance-v3-cache#123): payload execution countdowns use the real per-chain `delaySeconds`; every step links to its actual transaction (proposal `created/votingActivated/queued/executed/cancelled` + the voting-chain `closeAndSendVote`, and per-chain payload `created/queued/executed`); the "Voting closed" date uses `votingClosedAndSentAt`.
- **On-chain constant**: new `useGovernanceCoreConstants` hook reads `GovernanceCore.cooldownPeriod` once (cached, immutable) to drive the Queued→Executed timelock countdown. Only graceful fallbacks remain (payload delay if not yet snapshotted; the cooldown placeholder until the on-chain read resolves).

## Notes
- Payload `actions` depend on the cache backend exposing the field (already live on staging).
- The new timeline is the shipped proposal timeline (old component + dev preview page removed). Countdowns and links are driven by real data (cache fields + on-chain `cooldownPeriod`); only graceful fallbacks remain. Requires aave/governance-v3-cache#123 deployed for the payload/proposal fields.
- No new npm dependencies.


sammdec (dismissed):

sammdec:
Once this PR #3074 gets merged we can pull thsoe changes in here and the CI CD builds should start working again

sammdec (approved):

grothem (approved):

PR #3076, opened 2026-07-21
fix: crash on ENS names with disallowed unicode characters

mgrabina:
## Problem
The app white-screens for wallets whose primary ENS name contains characters that fail ENSIP-15 normalization (reported for 0xfb3c7eb936cAA12B5A884d612393969A557d4307, a zalgo-text .eth name). Two dependencies call viem's `normalize()` on the reverse-resolved name during React render with no guard, and the throw unmounts the whole tree:

1. connectkit@1.9.2, Avatar component (crashes the connect flow / account modal). Upstream fix proposed: https://github.com/family/connectkit/pull/513
2. @funkit/connect@9.28.1, useMainnetEnsAvatar hook (crashes at connect whenever funkit is configured, i.e. builds with NEXT_PUBLIC_FUNKIT_API_KEY set).

Reproducible by connecting as that address (e.g. via impersonator.xyz). Watch mode is unaffected because it never mounts either component.

## Fix
Wrap the `normalize()` call in try/catch in both packages, falling back to each package's existing no-ENS-name behavior (avatar query disabled, fallback avatar rendered, name still displayed raw). Both packages are on their latest release, so there is no fixed version to upgrade to.

Applied through pnpm's native `patchedDependencies` (patch files under `patches/`, referenced from `package.json` and pinned by hash in `pnpm-lock.yaml`). pnpm applies them at install time, so `pnpm install --frozen-lockfile` on CI reproduces them from the committed hashes. Added `*.patch` to `.prettierignore` since prettier has no parser for patch files and the pre-commit hook fails on them.

Note: this replaces the original patch-package approach after main migrated to pnpm; the funkit patch is regenerated against 9.28.1 (main's current version), where the vulnerable call site is unchanged.

## Verification
- Both patches apply on a clean `pnpm install --frozen-lockfile` (node 24 / pnpm 10.18) and the try/catch is present in the installed `connectkit` and `@funkit/connect` files.
- The wrapped code is behavior-identical to the original fix, which was verified against the reported wallet: reproduced the connectkit crash locally (EIP-6963 provider impersonating the address, open account modal) and traced the remaining funkit crash through the deployed sourcemaps to `useMainnetEnsAvatar`; after the patch, connect / account modal / dashboard / markets / governance / history all render with no `Invalid label` errors.
- Swept every copy of ens-normalize in the dependency tree (searched by its internal error strings) — these two are the only render-time call sites fed the wallet's reverse-resolved name.
- `pnpm lint:formatting` passes and patch files are correctly ignored.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -27,4 +27,5 @@ src/locales/
 *.mermaid
 *.log
 *.lock
+*.patch
 pnpm-lock.yaml
--- a/package.json
+++ b/package.json
@@ -20,6 +20,10 @@
       "form-data": "2.5.6",
       "d3-color": "^3.1.0",
       "js-cookie": "3.0.7"
+    },
+    "patchedDependencies": {
+      "connectkit@1.9.2": "patches/connectkit@1.9.2.patch",
+      "@funkit/connect@9.28.1": "patches/@funkit__connect@9.28.1.patch"
     }
   },
   "scripts": {
--- a/patches/@funkit__connect@9.28.1.patch
+++ b/patches/@funkit__connect@9.28.1.patch
@@ -1,0 +1,13 @@
+diff --git a/dist/index.js b/dist/index.js
+index 06a803f781f4a101d2bdefd2f398f4387a3d9bfa..2cc34b47f37878b1e04d83e545248d35208290e5 100644
+--- a/dist/index.js
++++ b/dist/index.js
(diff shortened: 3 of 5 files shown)


sammdec (dismissed):

sammdec:
@lochie FYI this is related to ConnectKit

sammdec (dismissed):

sammdec (approved):

JoaquinBattilana (approved):

PR #3084, opened 2026-07-28
fix: disable Polygon v3 incentives (UI_INCENTIVE_DATA_PROVIDER reverting)

grothem:
## Problem

`UI_INCENTIVE_DATA_PROVIDER` on Polygon v3 is reverting because the oracle for one of the reward tokens reverts. Since `usePoolsFormattedReserves` combines the reserves and incentives queries via `combineQueries`, an error in the incentives query leaves the combined result with `data: undefined` — taking down the whole Polygon market view, not just the incentives column.

## Fix

Disable the on-chain incentives path for Polygon v3 entirely. There are no rewards currently configured on the market, so there is nothing to display.

### DIFF
--- a/src/ui-config/marketsConfig.tsx
+++ b/src/ui-config/marketsConfig.tsx
@@ -494,7 +494,7 @@ export const marketsData: {
     v3: true,
     enabledFeatures: {
       liquiditySwap: true,
-      incentives: true,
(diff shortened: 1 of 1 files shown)


mgrabina (approved):
wondering if a trycatch on usePoolsFormattedReserves may help

JoaquinBattilana (approved):

PR #3090, opened 2026-08-18
feat: let users cap their repay approval, and show the amount needed

sammdec:
## General Changes

- Adds an **Allowance** control to the Repay modal offering **exact amount** or **unlimited**, and puts the exact figure on screen
- Fixes `addTransaction` recording `MAX_UINT_AMOUNT` for every approval regardless of what was actually approved

## Background

From a support report on Aave V3 Ethereum (cbBTC, wallet `0xCA68…487a`). The user capped their repay approval and the UI would not accept it, prompting for approval over and over. They eventually approved unlimited, repaid, then revoked.

Two independent problems produced that:

**1. There was no way to approve a finite amount.** `RepayActions` never passed `amountToApprove`, so [`generateApproval`](https://github.com/aave/interface/blob/main/src/store/poolSlice.ts#L263) fell through to `MAX_UINT_AMOUNT`. cbBTC's underlying token isn't in `permitConfig` for mainnet (only `acbBTC` is), so `tryPermit` was false and the "Approve with" cog never rendered either. The user was capping the spend limit by hand in their wallet because the app offered nothing else.

**2. A full repay demands more than the debt, and never says so.** The gate compares the allowance against `debt * 1.0025`:

| | |
|---|---|
| Debt at block 25703709 | `32.68074616` |
| Approved by the user | `32.70000000` |
| What the gate wanted | `32.76244803` |
| What the repay actually pulled | `32.68074630` |

Their approval was comfortably sufficient on-chain — the repay that eventually landed moved less than they had approved. Only the frontend rejected it, and it never showed the number it was asking for.

## Approach

The control is Repay's own (`RepayAllowanceControl`) rather than an option on the shared `ApprovalMethodToggleButton`. That component is generic across seven flows, so it has no access to the amount — and the amount is the whole problem. A Repay-owned control can render `Allowance 32.76244803 cbBTC` in the footer, so the transparency fix ships with the finite-approval option instead of trailing it as separate work.

`ApprovalMethodToggleButton`, `RightHelperText` and `walletSlice` are untouched. `TxActionsWrapper` gains an `approvalOptions` slot; no other flow changes behaviour and the existing Cypress selectors still work.

The choice is component state rather than a persisted preference — it's made inside a modal the user is already in.

## Developer Notes

**The margin is load-bearing, please don't remove it.** `maxApproveNeeded` is derived from live debt, so it creeps upward while an approval is in flight (~18 raw units/minute for this position). Approving *exactly* what the gate asked for at t0 falls short by t1 and recreates the reported loop using our own button. `getRepayAmountToApprove` applies `REPAY_ALL_APPROVAL_MARGIN` on top of `REPAY_ALL_BUFFER`, giving ~0.5% total on a full repay — weeks of accrual headroom. `src/components/transactions/__tests__/utils.test.ts` pins this, and the drift test fails if the margin is dropped.

Both the gate and the approval read from one `amountRequiringApproval` binding in `RepayActions`, so they can't diverge.

**Permit is untouched** — it already signs for an exact amount, so the control is hidden when permit is in use.

**Full precision is deliberate.** Rendering `32.7624` via `FormattedNumber` while approving `32.76244803` understates the allowance being granted, which defeats the purpose. Checked against an 18-decimal amount — it still fits the footer on one line.

**Drive-by:** `maxApproveNeeded` switches from `toString()` to `toString(10)`. It feeds `parseUnits` for the first time, which throws on the exponential notation BigNumber emits for dust-sized debts.

**Scope:** Repay only. Supply, Withdraw & Unwrap, Umbrella and the staking flows still approve unlimited. The `approvalOptions` slot is generic, so extending is straightforward if we want it.

**Design input welcome** on the "Allowance" wording and the menu's secondary lines — that copy hasn't had a designer look at it.

### Verification

- New unit suite passes (7 tests). The 4 pre-existing failing suites fail identically on `main` (ESM transform in `networksConfig`)
- `tsc --noEmit`, `eslint` and `pnpm build` all clean
- Control rendered and visually checked in the app's own theme, including an 18-decimal worst case. **Not** yet exercised against a fork — worth a manual repay pass before merge

---

## Reviewer Checklist


src/components/transactions/Repay/RepayActions.tsx
```diff
@@ -127,6 +143,9 @@ export const RepayActions = ({
     symbol,
     decimals: poolReserve.decimals,
     signatureAmount: amountToRepay,
+    amountToApprove: approveExactAmount
+      ? parseUnits(amountToApprove, poolReserve.decimals).toString()
```
mgrabina, line 147: This runs during render, and `getRepayAmountToApprove` returns the typed string untouched on the partial branch. The repay input has no `decimalScale` (`AssetInput.tsx:39`), so USDC (6dp) + `1.1234567` + "Exact amount" throws `fractional component exceeds decimals` and takes the modal down. The existing `parseUnits` calls are inside `action()`'s try/catch, so today the same input just fails the tx.

`.decimalPlaces(decimals, BigNumber.ROUND_UP).toString(10)` on both branches fixes it, and stops the footer showing more decimals than the token has.

src/components/transactions/utils.ts
```diff
@@ -20,6 +21,50 @@ export const useFlashloan = (healthFactor: string, hfEffectOfFromAmount: string)
...
+ */
+export const REPAY_ALL_BUFFER = '1.0025';
+
+/**
+ * Additional margin applied on top of `REPAY_ALL_BUFFER` when building the approval
+ * itself. `checkRequiresApproval` compares the allowance against a target derived from
+ * live debt, so that target creeps upward while the approval is in flight; approving
+ * exactly what the gate asked for would leave the user re-approving forever.
+ */
+export const REPAY_ALL_APPROVAL_MARGIN = '1.0025';
```
mgrabina, line 37: The margin leaves a residual allowance, which breaks USDT on Ethereum. Exact on a max repay approves `debt * 1.0025 * 1.0025` but only pulls the debt, so ~0.5% survives. USDT reverts `approve(spender, non-zero)` over a non-zero allowance, and the reset path returns early when `signatureAmount === '-1'` (`useApprovalTx.tsx:79`), which is the max-repay case. The next full repay fails at gas estimation.

Unlimited never hit this since `requiresApproval` stays false after one max approval. Either hide Exact for USDT-on-Ethereum, or key `needsUSDTApprovalReset` off the approval amount rather than `signatureAmount`.

src/hooks/useApprovalTx.tsx
```diff
@@ -207,7 +207,7 @@ export const useApprovalTx = ({
             action: ProtocolAction.approval,
             txState: 'success',
             asset: assetAddress,
-            amount: MAX_UINT_AMOUNT,
+            amount: amountToApprove ?? MAX_UINT_AMOUNT,
```
mgrabina: `amountToApprove` is base units, so the Amplitude `tokenAmount` on `GENERAL.TRANSACTION` will report `3276244803` next to the `32.68` every other caller sends (`RepayActions.tsx:209`, `SupplyActions.tsx:276`). `formatUnits(amountToApprove, decimals)` keeps it consistent.

Not repay-only either: Bridge, Umbrella stake and unstake already pass `amountToApprove`, so their approval events change shape too.

src/components/transactions/Repay/RepayActions.tsx
sammdec: Confirmed - `NumberFormatCustom` has no `decimalScale`, so 7dp on a 6dp token is reachable by typing, and this one runs at render rather than inside `action()`.

`getRepayAmountToApprove` now rounds on both branches (`.decimalPlaces(decimals, ROUND_UP).toString(10)`), so `parseUnits` cannot throw and the footer stops showing decimals the token does not have. Rounding up rather than down keeps the approval at or above the gate target, so no approve loop. Two tests added in `utils.test.ts` for the trim and for the gate still passing afterwards.

src/components/transactions/utils.ts
sammdec: Went with keying the reset off the approval amount, so Exact stays available for USDT.

`useApprovalTx` now derives `newApprovalAmount = amountToApprove ? formatUnits(amountToApprove, decimals) : signatureAmount` and the effect gates on that instead of `signatureAmount`, with the `0`/`-1` guards folded into `Number(newApprovalAmount) <= 0`. On a max repay with Exact selected that is a real finite figure, so `needsUSDTApprovalReset` sees the residual and the reset path runs.

Checked the other callers for fallout: Bridge and Umbrella unstake never pass `setShowUSDTResetWarning`, so they still return early, and Umbrella stake's `amountToApprove` is `parseUnits(signatureAmount)`, so it round-trips to the same value. The repay max + Exact case is the only behaviour change.

src/hooks/useApprovalTx.tsx
sammdec: Right - every other `addTransaction` caller passes token units (`RepayActions.tsx:228`, `SupplyActions.tsx:276`), so this would have been the odd one out.

Changed to `amountToApprove ? formatUnits(amountToApprove, decimals) : MAX_UINT_AMOUNT`, which also picks up Bridge and Umbrella stake/unstake since they were already passing `amountToApprove`.

mgrabina (dismissed):

grothem:
A couple of visual things I noticed when testing:

layout issue after selecting the exact amount option
image


on mobile, the text isn't wrapping
image

src/hooks/useApprovalTx.tsx
```diff
@@ -68,23 +68,29 @@ export const useApprovalTx = ({
 
   const [requiresApprovalReset, setRequiresApprovalReset] = useState(false);
 
+  // What the next approval will actually be for, in token units. `signatureAmount` is only
```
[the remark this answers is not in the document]
grothem: Accepted — the sentinel was doing double duty and `Number('-1') <= 0` swallowed it.

Rather than nest another ternary in the hook, the resolution now lives in `getNewApprovalAmount(amountToApprove, decimals, signatureAmount)` in `usdtHelpers.ts`: finite allowance wins, `'-1'` reads as `MAX_UINT_AMOUNT` (the approval it actually stands for), everything else passes through. Same behaviour as the suggestion.

Only Repay reaches the `'-1'` branch of the three callers that pass `setShowUSDTResetWarning` — Supply never sends the sentinel, and Umbrella stake always passes a finite `amountToApprove`.

Pinned in a new `src/utils/__tests__/usdtHelpers.test.ts` (8 tests), including the exact sequence here: residual allowance from an Exact repay, then a full repay with Unlimited, asserting the reset runs. This ternary has been wrong twice now, so it seemed worth locking down.

`71162ef`

grothem:
Both fixed in `f8890d6`.

**Layout after selecting Exact.** The footer row was `flex` with no wrap, so the long figure made the row squeeze instead of reflow — that is what split "Approve with" across two lines. The row now wraps, and `RepayAllowanceControl` has `flexShrink: 0`, so a long amount drops to its own line rather than compressing its neighbour. Short amounts (USDC `1250.5`) still share the line exactly as before. `maxWidth: 100%` caps the control at the row width, so an extreme amount breaks instead of overflowing the modal — the 18-decimal WETH case used to run off the card edge entirely.

**Mobile text not wrapping.** MUI's `MenuItem` is `white-space: nowrap` by default, which clipped the secondary line at the menu's right edge. Menu items now wrap, the paper is capped at `min(360px, calc(100vw - 32px))`, and the check icon is top-aligned so it sits with the option title rather than floating against a two-line description.

I checked this properly rather than by eye this time — a throwaway page rendering the real components in the app theme, screenshotted at both modal widths (420px and 359px), menu open and closed, across 6/8/18-decimal amounts. Your two screenshots reproduced exactly before the change and are clean after. The page is deleted, not committed.

Worth flagging: my PR note claimed the full-precision figure "still fits the footer on one line" against an 18-decimal amount. That was wrong — it only held without the permit toggle beside it, and GHO is exactly the case where both render.

`tsc --noEmit`, `eslint` and the unit suites are clean. Still not exercised against a fork, so the manual repay pass is still worth doing before merge.

mgrabina (dismissed):

JoaquinBattilana (dismissed):

mgrabina (approved):

JoaquinBattilana (approved):

